PR #4, opened 2016-03-25
Authorizes PRINT_EXPR after expr. Make single mode compile work

paradoxxxzero:
This makes single compile work:

```python
deparse_code(v, compile('int(1 + 1)', '<string>', 'single'))
```

Please review because I'm not really sure of what I did.

### DIFF
--- a/pytest/test_deparse.py
+++ b/pytest/test_deparse.py
@@ -313,7 +313,7 @@ def test_single_mode():
         'i = {}',
         'i = []',
         'while i < 1 or stop:\n    i\n',
-        'while i < 1 or stop:\n    print(i)\n',
+        'while i < 1 or stop:\n    print%s\n' % ('(i)' if PYTHON3 else ' i'),
         'for i in range(10):\n    i\n',
         'for i in range(10):\n    for j in range(10):\n        i + j\n',
         'try:\n    i\nexcept Exception:\n    j\nelse:\n    k\n'
--- a/pytest/test_deparse.py
+++ b/pytest/test_deparse.py
@@ -1,5 +1,5 @@
 from uncompyle6.semantics.fragments import deparse_code as deparse
-from uncompyle6 import PYTHON_VERSION, PYTHON3
+from uncompyle6 import PYTHON_VERSION, PYTHON3, deparse_code
 
 def map_stmts(x, y):
     x = []
@@ -302,3 +302,23 @@ for i in range(2): ...
     parsed = get_parsed_for_fn(for_range_stmt)
     if not PYTHON3:
         check_expect(expect, parsed)
+
+
+def test_single_mode():
+    single_expressions = (
+        'i = 1',
+        'i and (j or k)',
+        'i += 1',
+        'i = j % 4',
+        'i = {}',
+        'i = []',
+        'while i < 1 or stop:\n    i\n',
+        'while i < 1 or stop:\n    print(i)\n',
+        'for i in range(10):\n    i\n',
+        'for i in range(10):\n    for j in range(10):\n        i + j\n',
+        'try:\n    i\nexcept Exception:\n    j\nelse:\n    k\n'
+    )
+
+    for expr in single_expressions:
+        code = compile(expr + '\n', '<string>', 'single')
+        assert deparse_code(PYTHON_VERSION, code).text == expr + '\n'
--- a/uncompyle6/parsers/parse2.py
+++ b/uncompyle6/parsers/parse2.py
@@ -285,6 +285,7 @@ class Python2Parser(PythonParser):
         stmt ::= classdef
         stmt ::= call_stmt
         call_stmt ::= expr POP_TOP
+        call_stmt ::= expr PRINT_EXPR
 
         stmt ::= return_stmt
         return_stmt ::= ret_expr RETURN_VALUE
--- a/uncompyle6/parsers/parse3.py
+++ b/uncompyle6/parsers/parse3.py
@@ -302,6 +302,7 @@ class Python3Parser(PythonParser):
         stmt ::= classdef
         stmt ::= call_stmt
         call_stmt ::= expr POP_TOP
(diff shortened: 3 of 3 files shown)


rocky:
Sorry for the delay. I have been very busy.  

I am delighted that you take an interest here and have dug down into the weeds to do a pull request. In a superficial a way what you suggest makes sense. But I don't understand the details fully, so I can't pass judgement or verify yet.

Perhaps I'd understand what you did that causes a problem right now. For example showing  what you typed and what you got would be helpful. Writing a test for your change would also work, so I could see how it is broken now and is fixed after the change. 

The last things are minor. I think it would be helpful to more of the commentary that is currently above in the pull request in addition into the commit message for example:

     Make single-mode compile work:

     deparse_code(v, compile('int(1 + 1)', '<string>', 'single'))

Except as mentioned above, I'm not sure the relation of PRINT_EXPR to that above code. 

Finally you can add now comments to the grammar. See for example https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse2.py#L410-L413

paradoxxxzero:
Sorry for the delay too.
I'm sorry if my PR description was a little short, I will try to be more explicit.
I was working on my python debugger [wdb](/Kozea/wdb) and I wanted a way to get the source code from a frame object with no corresponding file, for instance when a command is evaluated from the interactive shell.

The python interpreter compiles interactive prompt input using the 'single' mode which compiles a single statement and add a print of non-None results. [doc](https://docs.python.org/3.6/library/functions.html#compile). This print is added with a PRINT_EXPR in byte code:
```python
>>> disco(compile('int(1 + 1)', '<string>', 'exec'))
  1           0 LOAD_NAME                0 (int)
              3 LOAD_CONST               2 (2)
              6 CALL_FUNCTION            1 (1 positional, 0 keyword pair)
              9 POP_TOP
             10 LOAD_CONST               1 (None)
             13 RETURN_VALUE
>>> disco(compile('int(1 + 1)', '<string>', 'single'))
  1           0 LOAD_NAME                0 (int)
              3 LOAD_CONST               2 (2)
              6 CALL_FUNCTION            1 (1 positional, 0 keyword pair)
              9 PRINT_EXPR
             10 LOAD_CONST               1 (None)
             13 RETURN_VALUE
```
My change was just in order to authorize the PRINT_EXPR operation which apparently is not currently supported. I am not really sure of what I changed in the grammar though, I just mimicked the POP_TOP behaviour.

rocky:
Ah - Now I understand where you are coming from. Thanks.

It makes much more sense start out with understanding of the problem, before judging the merits of a proposed solution to the unstated problem. Tell your friends ;-)

So it seems to me that this isn't a problem in parsing well-formed Python programs or modules, so much as wanted to support a slightly different kind of Python: the grammar used in "single-mode" compilation. 

If that's the case, then why not offer a flag on uncompyle6 to indicate that the grammar we want to be used in deparsing is the one that is used in single-mode compilation?  That way, instead of one parser that is really the merger of two grammars, we can have a two parsers that reflects the specifics of each grammar. It also makes it easier to comprehend each grammar, debug it, and it is less prone to unintended bugs caused by one deparse grammar bleeding into the other.

If you are on board with this, then I think the next step is to come up with the goal non-terminal used in single-mode compilation. I'm sure it's buried somewhere in the Python code. 

BTW good luck in wdb! You might also consider using uncompyle6 to flag exactly *where* in the line you are stopped at. I do that now in the trepan debugger, but I think this might be easily adapted here as well.

rocky:
@paradoxxxzero if you are starting with an expression rather than a full module, exactly how do you call deparsing? I imagine most people start by passing a byte-compiled file, but that can't be happening here since the byte-compiled file works at the module level.

rocky:
From CPython source Grammar/Grammar: 

```
# Start symbols for the grammar:
#       single_input is a single interactive statement;
#       file_input is a module or sequence of commands read from an input file;
#       eval_input is the input for the eval() and input() functions.
# NB: compound_stmt in single_input is followed by extra NEWLINE!
single_input: NEWLINE | simple_stmt | compound_stmt NEWLINE
file_input: (NEWLINE | stmt)* ENDMARKER
eval_input: testlist NEWLINE* ENDMARKER
```

paradoxxxzero:
I agree with you that it's two different grammars, but they are 99.99% similar and don't seem to overlap (PRINT_EXPR seems to be only used in this very case https://github.com/python/cpython/blob/master/Python/compile.c#L2597) so I don't know if it's worth to make a whole new parser.

I'm not sure I understand your second question but for now deparsing in wdb is done only here https://github.com/Kozea/wdb/blob/master/client/wdb/utils.py#L31 and I agree this is not the classic use case for uncompyle.

rocky:
Programming languages have a way of having code that largely the same, no? I've heard that there are these things called "if" statements and "subroutines".  

Clarity and modularity are a good things. If something is different, it is good idea to have that distinction be reflected in the code. 

The link to the use in wdb, clears up my question. Basically I'm asking how you would write a test case for this new code, since you can do it by reading a byte-compiled file. 

Instead you would have to do it via calling something like the function below: 

```python
def get_source_from_byte_code(code):
    version = sys.version_info.major + (sys.version_info.minor / 10.0)
    return uncompyle6.deparse_code(version, code).text
```
So would you write a test case which covers the grammar change you suggest?

paradoxxxzero:
Okay I added a very high level test. This is not ideal but gives an idea of what the problem is. As for the second grammar I'm still not sure, I mean nothing strictly forbid regular byte-code to use PRINT_EXPR and to me it's the same "byte-code grammar" just a different branch.

rocky:
Branch single-compile represents my current thinking on how to add this in merging in the code. I haven't added the additional grammar rule, but it would go inside class Python{2,3}ParserSingle

rocky:
The reworked code is now in the master branch.

PR #15, opened 2016-05-13
Fix for class decorators

Tey:
This PR fixes the decompilation of class decorators. It crashes with current master version because the handler for `classdefdeco2` is the same as for `classdef` while the structure of both nodes is slightly different.

The changes simply imply checking the node type and choose the appropriate child to consider as the `buildclass`. Also, there are some cosmetic changes so that the decorations are glued to the class definition and glued together.

You can test the changes by compiling [this Python file](https://gist.github.com/Tey/0d00c9313ab6770ff211328552d411e1#file-classdec_test-py).

**Similar changes are probably needed for Python 3 files** ([line 1142](https://github.com/rocky/python-uncompyle6/pull/15/files#diff-2f08c9a8823cc79320bcf701cf255fc9R1142)), but I don't have such a version installed so I can't test it.


rocky:
Thanks. See future commits for how to add this as a test and the changes to your file to make it Python3 compatible.

PR #19, opened 2016-05-25
Global update and bug fixing for Python 2.7.x 

StranikS-Scan:


rocky:
Hey, I am getting a sense that the pull requests you are thinking about are going impossible for a human to review, so let me set some guidelines. 

If the changes to existing files is more than say 100 lines, then you probably need to break up the pull request. 

Don't try in one pull request to fix 10 or so bugs in one pull request. Make one pull request per bug. 

Thanks.

PR #20, opened 2016-05-25
Global update and bug fixing for Python 2.7.x

StranikS-Scan:


rocky:
There are merge conflicts. Also automated tests are probably not going to work. 

This pull request is an order of magnitude too large. 

Create a pull request for each bug fixed so it is possible for me to understand what's going on.

PR #39, opened 2016-07-24
Pypy

rocky:
@moagstar please review. I'll be merging this into master even though there needs to be more work on pypy 3.2 and pypy 2.7 .

The last day has been hard to work on as a result of your and my changes in xdis. 

So instead of more rebases and merges, I'll be putting what's here into master after I fix the merge conflicts. 

There was a new release of xdis with your 3.6 changes and some pypy changes. However there will need to be another one to get on pypy 3.2 opcodes.


uncompyle6/scanners/scanner2.py
```diff
@@ -271,7 +276,7 @@ def build_lines_data(self, co, n):
 
         # linestarts is a tuple of (offset, line number).
         # Turn that in a has that we can index
-        self.linestarts = list(findlinestarts(co))
+        self.linestarts = list(self.opc.findlinestarts(co))
```
moagstar: In my work for python 3.6 support in uncompyle I have done the same thing, however for me the python2 bytecode tests are failing when running on python3 since the unmarshalling code is converting what were strings in python2 to bytes (findlinestrings for python2 calls ord). In the previous version of xdis findlinestarts was used from the standard library so if you were running python3 it made sense to perform this step in the unmarhsalling. Now xdis uses findlinestarts based on the version of bytecode being decompiled. I think some change will be required either here or in the unmarshalling to get this working correctly, @rocky do you have any thoughts on this?

rocky:
I thout the Python 3 interpreter working on Python2 bytecode I handled yesterday in merging your code with https://github.com/rocky/python-xdis/blob/master/xdis/bytecode.py#L28-L33

I am not seeing the failure you mention, nor is CircleCI: https://travis-ci.org/rocky/python-uncompyle6/jobs/146961046#L215-L245

So maybe if we're lucky all you need to do is merge in the existing  master branch of python-xdis and python-uncompyle6 code into your code base.

moagstar:
Ah that's true I will try first merging your changes.

PR #41, opened 2016-08-01
Move forward on moagstar's Python 3.6 support 

rocky:
@moagstar this is not the full spec. It doesn't handle for example format specifiers and `!r`, `!a` and so on. But I did want to move forward on Python 3.6. 

If you are interested and up for it, perhaps you would finish the job?


moagstar:
@rocky sure, thanks for making the start!

PR #43, opened 2016-08-11
Cache pip installation in travis

thedrow:


rocky:
https://docs.travis-ci.com/user/caching/ says that this is an experimental feature still. 

Caching is just for performance. Seems like a weird thing to want to change. 

I'd be grateful for an elaboration on why you felt it important for this change.

thedrow:
Faster builds are always important. They provide a faster feedback loop on your changes.
I know many repos that added caching.

If you don't want this, feel free to close.

rocky:
We can try this to see how much faster it goes. Another concern I had was when there is a problem in contributors code how fast they respond. That kind of speed is also important too and it looks like we're talking about on the order of 3 days or so.

rocky:
And for the record, this change did not speed up overall time at all. If anything it made it slower.

See https://travis-ci.org/rocky/python-uncompyle6/builds

PR #44, opened 2016-08-11
Test with latest PyPy in Travis

thedrow:


rocky:
I like the idea of testing on Pypy. I looks like right now there is an error using travis though.

I wonder how fragile adding custom code is going to be.

thedrow:
Sorry about the indentation issue. This should run now.

rocky:
It looks like there are two problems, uncovered here. First is the fact on pypy we shouldn't be verifying off of equivalent-version CPython. The second problem looks like there is a legitimate bug in deparsing on PyPy that should be fixed. The offending bytecode file has name `bsddb/dbrecio.pyc`

rocky:
Merged by accident - meant to keep this open.

PR #49, opened 2016-08-16
Fixed a bug with FORMAT_VALUE with sub expressions.

moagstar:
I have changed the processing of FORMAT_VALUE to allow for sub expressions. e.g. f'{len(items)}{x}'

I also added a hypothesis test which generates fstrings for testing. At the moment a couple of things are disabled in the test since these are known bugs that I still need to fix:

* conversion specifiers (!s, !r, !a)
* single expressions, e.g. f'{abc}'


rocky:
Great work! I love the fact that you added a test. 

I have been putting somewhat version-specific tests as bugxxx, but perhaps bytecode_xxx is more accurate. However in some cases and I suppose to some extend you could argue this here, this is half bytecode and have language feature. 

If you come up with a good naming scheme for classifying tests according to working from python versions x .. y (where y can be open ended), feel free to reorganize the files under test.

PR #56, opened 2016-09-21
merge fstring changes from moagstar

moagstar:
Changed fstring decompilation so that the call to join is left intact rather than trying to reconstruct the original fstring from the various expression parts. This allows f strings with 0, 1 and multiple expressions to be decompiled. It also simplifies decompilation of conversion specifiers !s !r !


moagstar:
@rocky do you want to review these changes before merging?

PR #59, opened 2016-10-21
Support for BUILD_STRING opcode

moagstar:


uncompyle6/semantics/pysource.py
```diff
@@ -1892,6 +1894,9 @@ def n_fstring_expr(self, node):
         node.conversion = self.FSTRING_CONVERSION_MAP.get(node.data[1].attr, '')
         self.default(node)
 
+    def n_fstring_single(self, node):
+        return self.n_fstring_expr(node)
+
```
rocky: Let me pull the code and check.

moagstar: Thanks, any suggestions are welcome

rocky: I think I understand a little bit more about what's going wrong, but it will take me more time to work this out.

PR #78, opened 2017-01-11
fix bug : not generate all files when use "-ro"

jlugjb:
when use the args of "-ro  outdir inputdir", only the first file is generated, other files is covered.

### DIFF
--- a/uncompyle6/main.py
+++ b/uncompyle6/main.py
@@ -143,6 +143,7 @@ def main(in_base, out_base, files, codes, outfile=None,
         # Try to uncompile the input file
         try:
             decompile_file(infile, outstream, showasm, showast, showgrammar)
+            outfile = None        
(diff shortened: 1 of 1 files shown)


rocky:
Thanks!

rocky:
With this change, in the tests when we run over checking all of the byte files I'm getting extra message regarding the files that are decompiled. 

I don't have time right to investigate what's up here. Yes, there is something unusual in that the same output file used over and over again (with the name of the first file). Using the same output file over and over again is fine for testing, but as you point out not good if you actually want to save all of those files. 

I'm guessing that by both concerns can be had by adding another option. But as I said, I don't have time to investigate how to do. So for now, I'm keeping things the old buggy (for your purposes) way.

Feel free though to work on this and come up with something that makes `make test` not be so noisy.

PR #80, opened 2017-01-18
Build const key map

moagstar:
Two changes in this branch, validation arbitrary code snippet decompilation (still needs a bit of work but works ok for simple expressions) , and support for python 3.6 BUILD_CONST_KEY_MAP.


rocky:
A cursory scan and this looks great. Thanks for doing. I'll probably look out and try in more detail over the weekend.

PR #81, opened 2017-01-19
fixed bug with BUILD_CONST_KEY_MAP

moagstar:
Missed this change from my last pull request which is necessary for the correct decompilation of BUILD_CONST_KEY_MAP


rocky:
ok. Thanks.

PR #84, opened 2017-03-04
Property based test function call

moagstar:
I've added some basic tests which should hopefully cover the majority of the new python 3.6 bytecodes and semantic changes for function calls. The majority of these tests are currently failing so I have marked these tests as expected failures. I wanted to merge these tests separately since I'd like to try and tackle the various issues in separate branches as I get time to do so, all of these will depend on these tests.

There is also a property based test which should test the various combinations more thoroughly, this test is skipped until the others are all passing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 /unpyc
 __pycache__
 build
+/.venv*
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@
 __pycache__
 build
 /.venv*
+/.idea
--- a/pytest/test_function_call.py
+++ b/pytest/test_function_call.py
@@ -1,7 +1,7 @@
 # std
 import string
 # 3rd party
-from hypothesis import given, assume, example, strategies as st
+from hypothesis import given, assume, settings, example, strategies as st
 import pytest
 # uncompyle
 from validate import validate_uncompyle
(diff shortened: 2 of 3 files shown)


rocky:
Thanks for undertaking this. It looks like in doing just the much you (and hypothesis) have uncovered at least one bug in the Python 3.x: 
```
def fn(a, b, d):
    return (a, b, d)

a = 10
b = {'b': 1,
     'd': 2}
fn(a=0, **b)
```

which is found in the 3.4 CircleCI test. The bug when using Python 3.3 is a mystery to me since it is saying the dis module doesn't have a Bytecode attribute but when I try it with 3.3.6 I don't have a problem. I'll try running this with Python 3.3.5 like the test does though.

moagstar:
I think I should change the usage of dis to xdis so that iterating the instructions works across all Python versions. Then these tests will be useful across the board not just in the later Python versions. I'll take a look.

rocky:
Sure, we can change to xdis. If you'd do that  I'd appreciate. 

I will have a bug fix for the hypothesis generated bug soon.

rocky:
I just realized that while using xdis may help with one of the CI failures here, it still doesn't fully help with verify since for best results you need to compile with a version of Python that closely matches the original bytecode. 

BTW the bug identified by hypothesis has been fixed.
Still, I guess it can't hurt.

rocky:
Many thanks!

moagstar:
I changed to using xdis for cross version python support. I should point out that the hypothesis validation is somewhat simpler than the other bytecode tests. The verification is performed by decompiling and disassembling a code snippet to test. The resulting decompiled source code is also disassembled, the bytecode from this step is compared with the raw bytecode from disassembling the code snippet (this comparison step needs improving, but for simple code snippets it works ok).

This means that we cannot verify that bytecode generated for example from python 3.6 can also be decompiled in 2.7. However it does verify that bytecode generated in 3.6 can be decompiled in 3.6. I have some ideas about how I could expand this to support also cross version verification. But for now I think that this can be put to good use for uncovering bugs and the main aim is to assist with python 3.6 support.

rocky:
> I should point out that the hypothesis validation is somewhat simpler than the other bytecode tests. 

I have been trying to understand what this means. I assume you are referring to the tests in say `test/bytecode_3.6`. 

But what is meant by _simpler_ ?  It uses less mechanism? Because it uses a function call rather than having to come up with a specific Python program, then byte-compile it using a specific version of Python and saving that to disk? And finally using the full `uncompyle6` program to run it? 

If that's the case, yes it is simpler.  And hypothesis tests can be smaller and more pin-pointable.

But there is always an implied point of view that "simpler is better".  You know, generally less moving parts means easier to understand and fix.

Quick, fast and "simple" tests are good. However I am not totally sure we'll ever be able to totally get rid of the tests in `test/bytecode_xx` for a number of reasons. First, when people report bugs, they come from concrete programs, and it may be easier initially to work off that rather than figure out how to express this in hypotheses. So in that sense, starting from a large Python program that exhibits a bug, and isolating it to a small region it is "simpler" or easy to do than the non-hypothesis way which the way I described it would take an additional step. 

Also, I should say that right now finding bugs in uncompyle6, is still a little bit too easy because there are still too many of them. Before working on the bug uncovered today I was working on another set of bugs in the bug queue. The bug is yet again control flow in Python 3.x in try/else statements. I imagine I'll work on that more the next free time I have on this.

So if we had a full blown hypothesis testing, we'd have a number of bugs that we'd just have to get through.

PR #113, opened 2017-05-03
Fixed out_base bug

grkov90:
Variable filename using in for

tags 
uncompyle6 -o haven't worked
argument -o haven't worked

### DIFF
--- a/uncompyle6/main.py
+++ b/uncompyle6/main.py
@@ -106,6 +106,7 @@ def main(in_base, out_base, files, codes, outfile=None,
         return open(outfile, 'w')
 
     tot_files = okay_files = failed_files = verify_failed_files = 0
+    current_outfile = outfile
 
     for filename in files:
         infile = os.path.join(in_base, filename)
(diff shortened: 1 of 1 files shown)


moagstar:
@grkov90 what was the bug that this pull request was supposed to fix? It seems that you renamed ``outfile`` to ``outfileindir``, but didn't then use ``outfileindir``. The tests are also failing, so perhaps you could take a further look at what is actually needed here?

grkov90:
Sorry, forgot to insert in 
`outstream = _get_outstream(outfileindir)`

rocky:
I am aware that there are a lot of little brokeness in the CLI. (And if I didn't care about backward compatibility I'd redo using click.) I don't understand though what _exactly_ is broken that you are referring to. I ran and got: 

```
$ uncompyle6 -o /tmp /usr/lib/python3/dist-packages/Crypto/__pycache__/__init__.cpython-35.pyc
# Successfully decompiled file
$ ls /tmp/__init__.cpython-35.py
/tmp/__init__.cpython-35.py
```
That's the level of detail that is useful.

grkov90:
Please try -o -r

grkov90:
You case work for one file. For multiple files work incorrect. Makes one file only

rocky:
This does correct the problem when options `-o -r` are used but unfortunately it brakes make test: 

```
$ make check
make[1]: Entering directory '/tmp/python-uncompyle6'
make -C test check-3.5
make[2]: Entering directory '/tmp/python-uncompyle6/test'
python test_pythonlib.py --bytecode-3.0 \
       --bytecode-3.1 --bytecode-3.2 --bytecode-3.3 \
       --bytecode-3.4 --bytecode-3.5 --bytecode-pypy3.2
Wed May  3 14:47:58 2017
Source directory:  bytecode_3.0
Output directory:  /tmp/py-dis-5brmaxog/bytecode_3.0

# okay decompiling bytecode_3.0/00_assign.pyc

# okay decompiling bytecode_3.0/00_import.pyc

...
### uncompile successful, but no file to compare against

### uncompile successful, but no file to compare against

### uncompile successful, but no file to compare against
```

There are the extra newlines, but more serious is those final lines about "no file to compare agains". Perhaps you can modify the code so both needs are satified?

rocky:
Thanks for taking the challenge and fixing that other code as well. 

Yes, this code is ugly and messy and not germain to the problem you specifically had. Someday all of this code will get cleaned up.

PR #118, opened 2017-06-09
scanners/scanner26: fixed undefined variable

AnythingTechPro:


rocky:
Thanks for the pull request. How about pull request #119 instead? It cleans up code just a tad and boy could this code use cleaning.

PR #160, opened 2018-02-24
Keep global statements in fixed order

wangym5106:
The `global` statements stored in a set() when generating code. The order of multiple `global` is not fixed.
When using a decompile bot to detect code changes, this will cause false positive.
This PR is to keep multiple `global` statements in fixed order.


rocky:
Sure, makes sense. 

 > When using a decompile bot to detect code changes, this will cause false positive.

Hmm... sounds interesting. Where can I get more information on this decompile bot?

wangym5106:
The decompile bot is just a script that downloads the updated version of a software, decompiles using uncompyle6, then check the diff with the previous version.

PR #223, opened 2019-04-14
Fixed 36 issues of type: PYTHON_E225 throughout 21 files in repo.

deepcommit:
PYTHON_E225: 'Fix extraneous whitespace around keywords.'.  This is a pep8 error code.          See <a href='https://pep8.readthedocs.io/en/latest/intro.html#error-codes'>        here for a complete list of error codes</a>.

It was fixed with <a href='https://github.com/hhatto/autopep8'>autopep8</a>.  The fix is completely safe.


rocky:
9b550b9d should now address.

> The fix is completely safe.
Yes, but it did mess up a concern. I've added a `# noqa` in the file so as to be more explicit. 

 A challenge with such large non-functional changes is that as we have subtle things as mentioned. It would be easy to slip in a bug in such massive changes.

There are many many many bugs in this code base that cause decompilation prolblems. Can I interest you in addressing one of those instead in the future?

Issue #227, opened 2019-05-09
args not appearing in decompiled src when kwargs is specified explicitly (call_ex_kw)


x0ret:
First of all, i should thank you for such an incredible efforts and educational project.

## Description

Suppose the following code:

```
def func():
    return testobj.testmethod(c, test="A", **extra_args)
```

using `uncompyle6` decompiles this to :

```
def func():
    return testobj.testmethod(test='A', **extra_args)
```

According to `customize36.py` this can be fixed with following patch:

```diff
diff --git a/uncompyle6/semantics/customize36.py b/uncompyle6/semantics/customize36.py
index 764a87e8..7e08c517 100644
--- a/uncompyle6/semantics/customize36.py
+++ b/uncompyle6/semantics/customize36.py
@@ -69,8 +69,8 @@ def customize_for_version36(self, version):
             '%c(%p)',
             (0, 'expr'), (1, 100)),
         'call_ex_kw' : (
-            '%c(%p)',
-            (0, 'expr'), (2, 100)),
+            '%c(*%c, %p)',
+            (0, 'expr'), (1, 'expr'), (2, 100)),

     })
```

## How to Reproduce

Please use following commands:

```bash
$ cat test5.py
def func():
    return testobj.testmethod(c, test="A", **extra_args)
$ python -m compileall .
Listing '.'...
Compiling './test5.py'..
$ uncompyle6 __pycache__/test5.cpython-36.pyc
# uncompyle6 version 3.3.2
# Python bytecode 3.6 (3379)
# Decompiled from: Python 3.6.7 (default, Oct 22 2018, 11:32:17)
# [GCC 8.2.0]
# Embedded file name: ./test5.py
# Compiled at: 2019-05-09 15:23:19
# Size of source mod 2**32: 81 bytes


def func():
    return testobj.testmethod(test='A', **extra_args)
# okay decompiling __pycache__/test5.cpython-36.pyc
```

## Expected behavior

decompilation like:

```python
def func():
    return testobj.testmethod(*(c,), test='A', **extra_args)
```

## Environment

- Uncompyle6 version: master HEAD `b57ca392`
- Python version: `3.6.7`
- OS and Version: Ubuntu bionic

Since i am new to this project if the above fix is accurate please let me know to propose a PR.

Thank you

rocky:
[...]

PR #230, opened 2019-05-09
implements n_call_ex_kw as discussed in #227

x0ret:
As discussed in #227, for a better pythonic form we need implementing `n_call_ex_kw` for have form:

```
return testobj.testmethod(c, test='A', **extra_args)
```
instead of:
```
return testobj.testmethod(*(c,), test='A', **extra_args)
```

### DIFF
--- a/uncompyle6/semantics/customize36.py
+++ b/uncompyle6/semantics/customize36.py
@@ -121,6 +121,32 @@ def customize_for_version36(self, version):
         return
     self.n_build_map_unpack_with_call = build_unpack_map_with_call
 
+    def call_ex_kw(node):
+        """Handle CALL_FUNCTION_EX 1 (have KW) but with
+        BUILD_MAP_UNPACK_WITH_CALL"""
+
+        expr = node[1]
+        assert expr == 'expr'
+        value = self.traverse(expr, indent='')
+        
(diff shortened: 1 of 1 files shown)


x0ret:
Lesson learned, use `python ./setup.py develop && COMPILE='--compile' make check` before sending PR.

rocky:
> Lesson learned, use python ./setup.py develop && COMPILE='--compile' make check before sending PR.

I do something different, but, I guess, equivalent.  I'm not trying to be sneaky or tricky here. If there is some place in the wiki or someplace to jot this down so others don't have the same problem, please update the wiki or suggest something else. Thanks.

x0ret:
Thanks for this informational comment.
I hope you like this commit coding style. Please adjust the PR as you wish or let me know to edit accordingly.

x0ret:
This comes to my mind so we can handle other function in this way? Is it correct? I mean by using `self.traverse` and `self.template_engine`.

I also observed we can implement this feature (args handling) in other methods like `call_ex_kw4`.

rocky:
Yes, I am sure there is lots of room for clean up and fixes in other code as well. Go at it!

I'll merge this now, but push a couple more small code cleanups and add the examples as tests.

PR #232, opened 2019-05-11
handling LOAD_CONST in build_tuple and multiple call_ex_kw(s) issues

x0ret:
1.
for some testcases like:

```python
f(2, 3, *b, *c, **a)
f(2, g, *b, *c, **a)
f(2, g, *b, *c, test='A', d=1)
f(2, 3, *b, *c, test='A', d=1)
```
which is handled by `call_ex_kw3` result was not correct.
also using template_engine instead of `self.write` and ... .

2. using template_engine in `call_ex_kw2` for better handling

3. fix LOAD_CONST and skiping first `*args ...` in `build_unpack_tuple_with_call`

4. normalizing `call_ex_kw4` positional_args.

@rocky, The main reason i'm sending this incomplete PR is to have advice from you to have better coding style and better handling cases. Please let me know whatever you think i should fix.

For some cases specially `call_ex_kw4` it's too hard for me to use template_engine. is it possible at all?

Thank you


x0ret:
next step we can form `call_ex` positional_args.

For now these are test cases i used :
```python
f(2, **a)
f(2, 3, *b, *c, **a)
f(2, g, *b, *c, **a)
f(2, g, *b, *c, test='A', d=1)
f(2, 3, *b, *c, test='A', d=1)
f((1, 2), 3, *b, *c, test='A', d=1)
f((1, 2), r, *b, *c, test='A', d=1)
f((x, y), r, *b, *c, test='A', d=1)
f(r, t, *b, *c, test='A', d=1)
f(*b, *c, test='A', d=1)
# kw4
f(2, r, **a)
# call_ex
f(2, *b)
f(2, *b, *c)
# kw2
f(2, g, *b, *c, test='A', d=1, **e)
```

I'm still trying to be comfortable with implementing test cases, I'll commit them later.

uncompyle6/semantics/customize36.py
```diff
@@ -148,61 +165,21 @@ def call_ex_kw2(node):
...
-        # This is weird shit. Thanks Python!
-        self.preorder(node[0])
-        self.write('(')
-
-        assert node[1] == 'build_tuple_unpack_with_call'
-        btuwc = node[1]
-        tup = btuwc[0]
-        if tup == 'expr':
-            tup = tup[0]
+        self.template_engine(
```
rocky: When a function has a single call to `self.template engine()`, followed by a call to `self.prune()` there is a shorthand for this by adding it into `TABLE_DIRECT` for example:
```
'call_ex_kw2': 
            ("%c(%p, %p)",
            (0, 'expr'), (1, 'build_tuple_unpack_with_call', 100),
            (2, 'build_map_unpack_with_call', 100)), node)
```

However the reason I didn't do that was probably because of handling the positional args like we do now in `call_ex_kw` now, thanks to you. 

So let's abstract the handling of the formatting of the positional arg like is done now in `call_ex_kw` and use that subroutine in `call_ex_kw2` now as well. What do you think?

x0ret: Yes, You are right, Please let me adjust as you said and retry.

x0ret:
> To add tests:

This is what i needed. Thank you.

I think i should prepare a wiki page for developing as you said before. I'll update so.

rocky:
> I  think i should prepare a wiki page for developing as you said before. I'll update so.

See, correct and add to: [Adding Tests](https://github.com/rocky/python-uncompyle6/wiki/Adding-Tests).

uncompyle6/semantics/customize36.py
x0ret: @rocky, there is one questions, in this commit i prepare `build_tuple_unpack_with_call` in the way it handles positional args formatting as we implemented before (For special case `LOAD_CONST`). the only thing is left if the result is empty we should omit `,`.  
besides if i get the concept right, in `call_ex_kw2` we have `build_tuple_unpack_with_call` but in `call_ex_kw` we don't have such an opr.  is it still needed to use our formatter? or possibly invoke `build_tuple_unpack_with_call` manually in `call_ex_kw2`?

Sorry for these questions. there's still a long way for me to understand all the concepts.
Thank you

rocky: I am not sure I know which is best. I leave it up to you. 

If there isn't much in the way of formatting, you do not have to call the formatter `template_engine()` . You can instead call `build_tuple_unpack_with_call()` directly if that is easier and clearer. 

I find all of this very complicated too. And that is also why the code is inconsistent and still a mess. 

If you are up to it after this 3.6 format strings is badly done. I have started code for that but haven't been able to finish it. Here's grammar rules really need to change.

x0ret: I see. i'll update this PR.

> If you are up to it after this 3.6 format strings is badly done. I have started code for that but haven't been able to finish it. Here's grammar rules really need to change.

 I prefer to help in order of your priorities. Say 3.6 fmt or whatever you decide.

x0ret:
I think we can merge and continue now. I'll review `cusotmize36.py` later. 

 I also added some test cases.

Thank you

rocky:
> I think we can merge and continue now. I'll review cusotmize36.py later.

Looks good to me too. Thanks for the fixes and cleanups.

I'll push what I have for format strings in its own branch a little later after  I reviewed it.  I believe some tests may fail in that branch.

PR #233, opened 2019-05-12
Revise format string handling

rocky:
@x0ret

Please go over this and fix. The main test program here is in [`test/bytecode_3.6_run/01_fstring.pyc`](https://github.com/rocky/python-uncompyle6/blob/master/test/bytecode_3.6_run/01_fstring.pyc).

Problems: 
- [x] there are extra parentheis in the format strings because precedece values are probaby not set or not set correctly. 
- [x] Check that python 3.7 `dataclasses.py` decompiles okay. It has a lot of string formatting in it, and this is how I found these problems
- [x] See if non-terminal names are okay and conform where possible to Python's [AST](https://docs.python.org/3/library/ast.html) terminology. Note that the two are never the same, because they can't be. So we use underscore instead of camelcase. And additional suffixes (like 1, 2, 36) are okay too. 
- [ ]  Go over the wilki entry [Adding a new language  construct: Python 3.6 Formatted String Literals](https://github.com/rocky/python-uncompyle6/wiki/Adding-a-new-language-construct:-Python-3.6-Formatted-String-Literals) and correct the mistakes there for what we now do or should do.

Of course there may be additional things you may want to change. If so, just go ahead and make the change.

Thanks!


x0ret:
Thank you for granting me into this.

about parenthesis issue, As you said i checked precedence values for following test case:
```python
source = f"__file__ = r'''{(os.path.abspath(filename))}'''\n" + source + '\ndel __file__'
```
and we have this:

```
n_expr::PRECEDENCE:  (kind=expr, prec=-2), (kind=formatted_value1, prec=-2)
```
So as you said since `formatted_value1`'s prec is not greater than `expr`'s prec we have parenthesis.
A simple attempt to set to a higher value fixed this behavior:
```diff
@@ -41,6 +41,8 @@ def customize_for_version36(self, version):
     PRECEDENCE['call_ex_kw3'] = 1
     PRECEDENCE['call_ex_kw4'] = 1
     PRECEDENCE['unmap_dict']  = 0
+    PRECEDENCE['formatted_value1'] = 100
+    PRECEDENCE['formatted_value2'] = 100
``` 

However i'm not sure about the value, since lower value than `10` is not enough as `expr` can be much more. Could you please comment on this?

rocky:
Everything you wrote is correct, so no (further) comment. Thanks.

rocky:
@x0ret please just make your commits on top of this. Thanks.

x0ret:
About python 3.7 `dataclasses.py` i think i checked all format strings we covered all the cases except one case: 
```python
def _repr_fn(fields):
    return _create_fn('__repr__',
                      ('self',),
                      ['return self.__class__.__qualname__ + f"(' +
                       ', '.join([f"{f.name}={{self.{f.name}!r}}"
                                  for f in fields]) +
                       ')"'])
```

which we had:

```python
'return self.__class__.__qualname__ + f"(' + ', '.join([f"{f.name}={self.{f.name}!r}" for f in fields]) + ')"'])
```

with the following fix i think we can handle these cases too:

```diff
diff --git a/uncompyle6/semantics/customize36.py b/uncompyle6/semantics/customize36.py
index 289ecc8a..fdebca8c 100644
--- a/uncompyle6/semantics/customize36.py
+++ b/uncompyle6/semantics/customize36.py
@@ -474,6 +474,8 @@ def customize_for_version36(self, version):
         for expr in node[:-1]:
             assert expr == 'expr'
             value = self.traverse(expr, indent='')
+            if not value.startswith("f"):
+                value = value.replace("{", "{{").replace("}", "}}")
             if expr[0].kind.startswith('formatted_value'):
                 # remove leading 'f'
                 value = value[1:]
```

after fix with a diff on output srcs we have:

```diff
@@ -224,7 +224,7 @@ def _init_fn(fields, frozen, has_post_in

 def _repr_fn(fields):
     return _create_fn('__repr__', ('self', ), [
-     'return self.__class__.__qualname__ + f"(' + ', '.join([f"{f.name}={self.{f.name}!r}" for f in fields]) + ')"'])
+     'return self.__class__.__qualname__ + f"(' + ', '.join([f"{f.name}={{self.{f.name}!r}}" for f in fields]) + ')"'])


 def _frozen_get_del_attr(cls, fields):
``` 

Could you please comment on this too?

rocky:
As I now understand the complexity indicated by `dataclasses.py` Inside the string part format strings  `{{` and `}}` become `{` and `}` respectively. But in the `LOAD_CONST` of the code, the single delimiter appears.

So instead of: 

```
             if not value.startswith("f"):
                 value = value.replace("{", "{{").replace("}", "}}")
             if expr[0].kind.startswith('formatted_value'):
                 # remove leading 'f'
                 value = value[1:]
```
```

             if expr[0].kind.startswith('formatted_value'):
                 # remove leading 'f'
                 value = value[1:]
             else:
                 assert expr[0] == 'LOAD_CONST'
                 value = value.replace("{", "{{").replace("}", "}}")
```
might be specific.  But we would _also_ need something similar inside `format_string1` and `format_string2` when they have constant string values in the non-formatting specifier `expr` subnode. 

A problem with testing on a leading `f` inside the value is that you can't from this alone if this was part of the string or not. The tree node however is unambiguous. 

- - - - 
The rest is a larger discussion around this topic. 

The Python 3.6 AST for this is: 
```
Module
  FunctionDef
    arguments
      arg
    Return
      Call
        Name
          Load
        Str
        Tuple
          Str
          Load
        List
          BinOp
            BinOp
              Str
              Add
              Call
                Attribute
                  Str
                  Load
                ListComp
                  JoinedStr
                    FormattedValue
                      Attribute
                        Name
                          Load
                        Load
                    Str
                    FormattedValue
                      Attribute
                        Name
                          Load
                        Load
                    Str
                  comprehension
                    Name
                      Store
                    Name
                      Load
            Add
            Str
          Load
```

In the AST `Str` is used and that is more precise than `LOAD_CONST`. Ideally just like we do to turn `LOAD_CONST` into `LOAD_ASSERT` sometimes in the ingest/scanner phase, it would be better if we added `LOAD_STR` and `LOAD_INT`, and `LOAD_TUPLE` for such specific constants. That we we could craft a more precise grammar what we have now say: 

```
expr ::= formatted_value1
expr ::= formated_value2
```
The `expr` would become `str` and this would better match the AST and become clearer. 

However this a lot of work.

x0ret:
> But we would also need something similar inside format_string1 and format_string2 when they have constant string values in the non-formatting specifier expr subnode.

I used following test case :
```python
f"{f}{'{name}'}"
```

`--tree+` results in : 
```
                         1. expr
                            formatted_value1 (2)
                                 0. expr
                                                 4  LOAD_CONST               '{name}'
                                 1.              6  FORMAT_VALUE          0  ''
                         2.              8  BUILD_STRING_2        2  ''
```

If we are talking about this kind of `LOAD_CONST` in subnode `expr` in `formatted_value1`, We should not replace `{` with `{{`. Am i correct?

rocky:
> If we are talking about this kind of LOAD_CONST in subnode expr in formatted_value1, We should not replace { with {{. Am i correct?

You are correct and I was wrong when I wrote: 

> But we would also need something similar inside format_string1 and format_string2 when they have constant string values in the non-formatting specifier expr subnode.

But this does point out why a blind string substitution without taking into regard the node type is wrong. The if/else code handles that while the test on the string just doesn't. 

I stress this because there's a lesson that I've learned here (and that I see many people make a mistake when doing this kind of stuff): working with strings is often expedient and simpler. In some cases it can be more general too. However in other cases working with strings can cause problems and often should be done in conjunction with understanding the tree nodes.

x0ret:
> But this does point out why a blind string substitution without taking into regard the node type is wrong. The if/else code handles that while the test on the string just doesn't.

You are right, my bad. I got your point.

> I stress this because there's a lesson that I've learned here (and that I see many people make a mistake when doing this kind of stuff): working with strings is often expedient and simpler. In some cases it can be more general too. However in other cases working with strings can cause problems and often should be done in conjunction with understanding the tree nodes.

I appreciate your sharing experience. I'll try to keep this in my mind.

For now i think we can have your fix as a commit and continue. 

Thank you

rocky:
@x0ret Ok working on it with additional tests...

Issue #235, opened 2019-05-13
Python 3.7 decompiling full 3.7.3 library without error


rocky:
## Description

```
$ pyenv local 3.7.3
$ cd test
$ ./test_pyenvlib.py --max 2200 --verify
```
Remove 36 or so parser errors. 

@x0ret as with the others, if this is not of interest or too much work, feel free to remove yourself.

x0ret:
```
  File "/Users/x0ret/Desktop/dev/python-uncompyle6/uncompyle6/semantics/pysource.py", line 1448, in print_super_classes3
    if i == start:
UnboundLocalError: local variable 'i' referenced before assignment
```

```python
                start = n-2
                for i in range(start, 0, -1):
                    if not node[i].kind in ['expr', 'call', 'LOAD_CLASSNAME']:
                        break
                    pass

                if i == start:
                    return
```

```diff
             else:
-                start = n-2
+                i = start = n-2
                 for i in range(start, 0, -1):
                     if not node[i].kind in ['expr', 'call', 'LOAD_CLASSNAME']:
```

Need more investigation. I'm working on it.

x0ret:
@rocky i was investigating python3.7 `datetime` issue and noticed the issue is another `chained compare`, this is the the test case:

```
def f():
    if (month is None and
	isinstance(year, (bytes, str)) and len(year) == 4 and
	1 <= ord(year[2:3]) <= 12):
        return 5
```

finally this is the diff:

```diff
diff --git a/uncompyle6/parsers/parse37.py b/uncompyle6/parsers/parse37.py
index 1d2baa7e..25c6fd28 100644
--- a/uncompyle6/parsers/parse37.py
+++ b/uncompyle6/parsers/parse37.py
@@ -99,6 +99,8 @@ class Python37Parser(Python36Parser):

         compare_chained1_false_37 ::= expr DUP_TOP ROT_THREE COMPARE_OP POP_JUMP_IF_FALSE
                                       compare_chained2c_37 POP_TOP JUMP_FORWARD COME_FROM
+        compare_chained1_false_37 ::= expr DUP_TOP ROT_THREE COMPARE_OP POP_JUMP_IF_FALSE
+                                      compare_chained2d_37 POP_TOP JUMP_FORWARD COME_FROM
         compare_chained2_false_37      ::= expr DUP_TOP ROT_THREE COMPARE_OP POP_JUMP_IF_FALSE
[...]

PR #236, opened 2019-05-13
Python 3.7 decompiling full 3.7.3 library without error #235

x0ret:
#235 

- fix a minor issue in pysource
- [x] fix chained compare parse error ending in `POP_TOP JUMP_BACK COME_FROM` (from `getopt.pyc`)
- [ ] issue in datetime parse error

I'm working on it.


rocky:
These look good - many thanks! 

Let's merge it and have lots of little PR's instead of one big one. BTW there is a small grammar typo in a comment I made that I am going to correct after merge.

For my curiosity, what example code in `print_super_classes3` was causing the range to be empty?

x0ret:
> For my curiosity, what example code in print_super_classes3 was causing the range to be empty?

This was before i read `test_pyenvlib.py` file which you ignore tests. but the file was 
`~/.pyenv/versions/3.7.3/lib/python3.7/test/__pycache__/test_enum.cpython-37.opt-2.pyc`. I didn't investigate further.

PR #237, opened 2019-05-14
Less dishonest COME_FROMs in 3.6+ code

rocky:
Addresses all of problems seen in 3.7 datetime.py

See also Issue #235

### DIFF
--- a/uncompyle6/parsers/parse36.py
+++ b/uncompyle6/parsers/parse36.py
@@ -33,6 +33,10 @@ class Python36Parser(Python35Parser):
         """
         sstmt ::= sstmt RETURN_LAST
 
+        cf_for_iter  ::= _come_froms FOR_ITER
+        list_for     ::= expr cf_for_iter store list_iter jb_or_c
+        genexpr_func ::= LOAD_FAST cf_for_iter store comp_iter JUMP_BACK
+
         # 3.6 redoes how return_closure works. FIXME: Isolate to LOAD_CLOSURE
         return_closure   ::= LOAD_CLOSURE DUP_TOP STORE_NAME RETURN_VALUE RETURN_LAST
 
--- a/uncompyle6/parsers/parse36.py
+++ b/uncompyle6/parsers/parse36.py
@@ -33,10 +33,6 @@ class Python36Parser(Python35Parser):
         """
         sstmt ::= sstmt RETURN_LAST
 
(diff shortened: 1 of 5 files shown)


x0ret:
Thank you, i think we can fix related issues in a reasonable time.
however i started reading [python-control-flow](https://github.com/rocky/python-control-flow). Maybe we can have a schedule to embed into this project.
 
I will check again the library and continue.

uncompyle6/parsers/parse36.py
```diff
@@ -33,6 +33,10 @@ def p_36misc(self, args):
         """
         sstmt ::= sstmt RETURN_LAST
 
+        cf_for_iter  ::= _come_froms FOR_ITER
```
rocky: Oops - this isn't needed (or used) anymore. The decision (for now) is not to add COME_FROMS to backward (in current Python code generation that means looping) jumps.

rocky:
I have thought a lot about using python-control-flow. It would require a very serious effort to get code as good as we have now - Python is complicated in its control flow. However it would revolutionize this project and put it a much better place for Python 3.7, 3.8 and beyond. And we could get rid of a lot of hacky code and things would be more intelligible. 

A strategy for rolling this in is to do it in a branch and just try one of the more recent Pythons on it. After that works expand to other Pythons. 

Another approach is to create an entirely new project that just has code for one Python that uses this. And after that works well expand it.

x0ret:
> A strategy for rolling this in is to do it in a branch and just try one of the more recent Pythons on it. After that works expand to other Pythons.

To be honest, I do prefer to have this progress in this project in the future : ) , since you do a lot of work on it. However i think isolating was always a good option for me. Maybe we can start new project for as you said (python 3.8?) and see the progress to make final decision. 

I'm trying to adjust myself to `python-control-flow` as  soon as possible.

I really appreciate your efforts in this project. This is so promising.

x0ret:
@rocky, i checked half of 3.7 lib issues, and it seems most of them are related to chained comparison, So do we continue in this PR?

rocky:
@x0ret I am going to stop working on this for today. I have too much other work. If you could get this to pass the test, I think we could have most of the 3.7 issues behind us. 

After that, maybe a release. And then focus more on control flow better way.

rocky:
Thanks! Let's do it and move on..

PR #243, opened 2019-05-27
Some docstring bugs fixed, some remain...

rocky:
I had broken escaping the tail quote by inadvertently switching from `"""`
by default to `'''`.

Some additional tests have been added to 00_docstring.py for
this. However...

Unicode decoding is still broken. For now I've added  `errors="ignore`  to
`.decode("utf-8", ...)` until a better fix is found. Sigh.

@x0ret If you can figure out how to fix this I'd be grateful.


rocky:
Oh - and this fixes CI that was broken in the last merge: https://travis-ci.org/rocky/python-uncompyle6/builds/537941263

x0ret:
> Unicode decoding is still broken. For now I've added errors="ignore" to
.decode("utf-8", ...) until a better fix is found. Sigh.

Could you please share the test case?

rocky:
> Could you please share the test case?

Sorry about that -  I was doing this on Python 2.7 since that is what was broken, If you look at the revised 00_docstirng.py I think that still covers this. And it has a test case commented out because I can't get that to work. Same thing in decompyle3 too.

test/simple_source/stmts/00_docstring.py
```diff
@@ -4,18 +4,42 @@
...
     r'''func placeholder - ' and with ("""\nstring\n""")'''
+    assert dq5.__doc__ == r'''func placeholder - ' and with ("""\nstring\n""")'''
 
-def bar():
+def dq6():
     r"""func placeholder - ' and with ('''\nstring\n''') and \"\"\"\nstring\n\"\"\" """
+    assert dq6.__doc__ == r"""func placeholder - ' and with ('''\nstring\n''') and \"\"\"\nstring\n\"\"\" """
+
+def dq7():
+  u"""        <----- SEE 'u' HERE
```
x0ret: These lines need more indentation same as others, otherwise test will fail.

rocky: Just to be clear. If you run this from python, I believe it doesn't give an error. It is just after decompiling. And decomping 3.7 byecode  works using unpy37. 

There is some fancy indentation code in uncompyle6 that unpy37 doesn't have. Perhaps that should be removed?

x0ret:
I'm sorry about that, Unfortunately i did not test with `-o` option and it seems most of the last merge changes can be reverted. However about this case:

```diff
         if self.version >= 2.4:
             if self.version > 2.7:
-                docstring = repr(docstring.expandtabs())[2:-1].decode("unicode-escape")
+                docstring = repr(docstring.expandtabs())[2:-1]
             else:
                 self.write('u')
-                docstring = repr(docstring.expandtabs())[2:-1].decode("string-escape")\
-                                                              .decode("utf-8", errors="ignore")
+                docstring = repr(docstring.expandtabs())[2:-1]
```

If it is possible, Please wait until i revise my last PR and fix these issues. 

Again, Sorry for this situation.

rocky:
I am happy to wait. Take your time.

test/simple_source/stmts/00_docstring.py
x0ret: Ok i checked `unpy37`, we use passed arg `indent` in `print_docstring` to print line-by-line. However the content of multiline docstring treat the string as raw:
```python
def t():
  """
test
  """
  assert t.__doc__ == """
test
  """

t()
```

is decompiled to :

```python
def t():
    """
    test
      """
    print(t.__doc__)
    if not t.__doc__ == '\ntest\n  ':
        raise AssertionError

t()
```

I believe we should not [parse](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/semantics/helper.py#L161) docstring and use docstring as is. then our indent is ok.

x0ret: This fixes.
```diff
@@ -184,12 +184,12 @@ def print_docstring(self, indent, docstring):
         self.println(trimmed[0])
         for line in trimmed[1:-1]:
             if line:
-                self.println( indent, line )
+                self.println( line )
             else:
                 self.println( "\n\n" )
                 pass
             pass
-        self.println(indent, trimmed[-1], quote)
+        self.println(trimmed[-1], quote)
```

rocky: Got it. Thanks!

rocky:
I've updated 00_docstring.py to reinstate a test, although I haven't redone the bytecode for 3.6 and 3.7. (It works on 2.7 oddly though).

test/simple_source/stmts/00_docstring.py
rocky: This change has been committed although, dq7 (unicode accents) fails on 3.{6,7}

x0ret: @rocky, I'm a little confused about [these](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/semantics/helper.py#L162-L172) lines. Why we need this?

Removing those lines and `trimmed = lines` (ugly) works in all versions as i checked. Could you please verify?

x0ret:
@rocky, Can i commit to this PR?

and please comment about this trick:
```python
        try:
            repr(docstring.expandtabs())[1:-1].encode("ascii")
        except UnicodeEncodeError:
            self.write('u')
```
Here i'm going to check whether string is unicode or not and precisely for Python2 i can use `u`, So we are not going to use chardet or ... . If it is ok to you, I'll commit changes.
All issues seem to be gone.

rocky:
> Here i'm going to check whether string is unicode or not and precisely for Python2 i can use u, So we are not going to use chardet or ... . If it is ok to you, I'll commit changes.

I tried the changes and I didn't see a problem. But then without the change I wasn't seeing a problem either. 

(If this the change were in a branch to merge into this one, it would save me cut and paste and the possibility of making an error). 

> All issues seem to be gone.

I was having problems with 3.{6,7} with the `baz` test in `00_docstring.py`. 

But feel free to merge this and other changes. We can always keep going if there are additional problems.

x0ret:
Please review and test [this branch](https://github.com/x0ret/python-uncompyle6/tree/docstrings-again).

rocky:
@x0ret That branch works great and looks good. Let's ship it!

rocky:
Superseded by #244

PR #244, opened 2019-05-28
Fix unicode docstring again

x0ret:
Fix unicode docstring again, handling unicode string in py2, fix docstring indentation


rocky:
Many thanks!

Issue #247, opened 2019-06-04
Extra parenthesis around in generators when they are the sole argument in a function call


rocky:
In Python 3.7ish

```
if all(str(exc) == model for exc in exceptions):
    raise exceptions[0]
```

has extra parenthesis:

```
if all((str(exc) == model for exc in exceptions)):
    raise exceptions[0]
```

x0ret:
As far as i can tell, The cause of this behavior is caused by these two routes:

1. [pysource.py:1020-1028](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/semantics/pysource.py#L1020-L1028) which will output inner parenthesis.
2. Second, [pysource.py:1986](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/semantics/pysource.py#L1986) which is responsible for handling call parenthesis.

Due to need information about inner `genexpr_exp`. I think we should fix this in handling call (access to parent node is possible?). otherwise we break tests like:
```
names = (formatmonthname(False) for k in __file__)
```

I'll investigate more.

x0ret:
Maybe something like this? 

```diff
diff --git a/uncompyle6/semantics/customize35.py b/uncompyle6/semantics/customize35.py
index 05639d5c..40df0477 100644
--- a/uncompyle6/semantics/customize35.py
+++ b/uncompyle6/semantics/customize35.py
@@ -113,6 +113,16 @@ def customize_for_version35(self, version):
                 self.template_engine(template, node)
             self.prune()

+        elif key.kind == "CALL_FUNCTION_1":
+            args_node = node[-2]
+            if args_node == 'pos_arg':
+                assert args_node[0] == 'expr'
+                n = args_node[0][0]
+                if n == 'generator_exp':
+                    template = ('%c%P', 0, (1, -1, ', ', 100))
+                    self.template_engine(template, node)
+                    self.prune()
+
         self.default(node)
     self.n_call = n_call
```

What do you think?

rocky:
The analysis is right on. Thanks for investigating. The fix is great too. 

I checked to see whether there could be functions other than `FUNCTION_1`, but I get the message `Generator functions must be parenthesized of not the sole argument`. 

So let's patch and ship it! I'd be grateful if you would handle and note that peculiarity which I am likely to forget (and others may be unaware of). 

In the end this seems like a small detail, which is why it is a _feature_ rather than a bug. But I think it cool to handle.

x0ret:
[...]

PR #248, opened 2019-06-05
feature #247: handle extra parenthesis in generators

x0ret:
As discussed in #247, for better formatting src, we are going to handle parenthesis in generators more accurate.

@rocky, as you can see due to `n_call` usage there were multiple cases to be changed. Is it ok to you?

### DIFF
--- a/uncompyle6/semantics/customize26_27.py
+++ b/uncompyle6/semantics/customize26_27.py
@@ -37,3 +37,22 @@ def customize_for_version26_27(self, version):
             'testtrue_then': ( 'not %p', (0, 22) ),
 
         })
+
+    def n_call(node):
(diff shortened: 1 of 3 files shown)


x0ret:
I checked following PY versions:

- 2.7
- 3.1.5
- 3.2.6
- 3.5 
- 3.7

rocky:
While this is very good, and in theory I have no problem with it. I thought I'd try to reduce the redundancy a little. It also shows a trick where we can put rules into the table and just change a node's type. That might also be slicker in tree display, but I don't know. 

Let me know what you think.

rocky:
Built upon this in #250. Many thanks for doing this though.

PR #252, opened 2019-06-07
[WIP] LOAD_CONST->LOAD_STR for Python 3.x

rocky:
@x0ret FYI 

In the future I also hope to see pushing back to Python 2.x. And also adding LOAD_CODE. Initial tests shows that this is more clear, is simpler, and speeds up parsing.


rocky:
Looks like I have tests to fix. Will address this over the weekend or later.

x0ret:
@rocky,
```diff
diff --git a/uncompyle6/semantics/consts.py b/uncompyle6/semantics/consts.py
index 3585920f..9ba0c0ee 100644
--- a/uncompyle6/semantics/consts.py
+++ b/uncompyle6/semantics/consts.py
@@ -318,7 +318,7 @@ TABLE_DIRECT = {
     'mkfuncdeco0':         ( '%|def %c\n', 0),
     'classdefdeco':        ( '\n\n%c', 0),
     'classdefdeco1':       ( '%|@%c\n%c', 0, 1),
-    'kwarg':               ( '%[0]{attr}=%c', 1),
+    'kwarg':               ( '%[0]{pattr}=%c', 1),
     'kwargs':              ( '%D', (0, maxint, ', ') ),
     'kwargs1':             ( '%D', (0, maxint, ', ') ),
```

fixes the CI issue. But i don't know it's cause. I hope it helps.

rocky:
@x0ret Thanks catching. In Python 2 we are still using `LOAD_CONST`. `attr` has its `str` value while `pattr` is the `repr()` value. For arbitrary constants which include sets and lists, we need to use repr()  which is fragile because we are relying on repr() to do the right thing in the context we need. 

However for strings we can avoid the fragility of `repr()`  and should be able to use `str()`. 

I am probably not explaining this well, and I forget all of the nuances, but using repr() in constances is prehaps another reason why `LOAD_CONST` should be made more specific as it already is. And another reason why probably `LOAD_CODE` in the future will be ever so slightly better.

rocky:
We may find more breakage, but we should try this.

Issue #169, opened 2018-04-01
Annotation args in 3.0 - 3.5


rocky:
We've punted on this for now.  3.6 on @x0ret has working.

DamlaAltun:
+1

rocky:
@DamlaAltun interested enough in trying your hand at this?

DamlaAltun:
:/ Sorry, i am not familiar with bytecode, parsing... Maybe Batuhan (@BTaskaya) can help you, he wrote a blog post about this bytecode tech.

rocky:
No one is born understanding bytecode parsing.

 I find it amusing that you are happy to volunteer someone else. Personally, though am not looking for help as I have no need of this feature right now.

rocky:
@x0ret We capture the information and even show return annotation information. Parameter information still isn't done. If this is of no interest, just unassign yourself.

rocky:
@x0ret Actually this may be fixed in 3.0-3.5, but there is a more general bug in parameter order. 

For Python 3.3 for 

```
def foo2(a, b: 'annotating b', c: int, *args: str) -> float:
``` 

we get: 

```
def foo2(*argsa, b: 'annotating b',
         c: int) -> float:
```

Python has switched around parameter ordering on the stack several times between versions.

x0ret:
Sorry, I admit i didn't check other versions. According to my investigation `make_function3_annotate` is responsible for such a behavior. 

> Python has switched around parameter ordering on the stack several times between versions.
I didn't know that. Thanks.

I'll check other versions and fix the issue. (I'm a little busy right now, will try tomorrow).

rocky:
> I'm a little busy right now, 

I totally understand. No problem. Thanks for all of the help.

PR #253, opened 2019-06-08
Revise annotation type implementation for < 3.6

x0ret:
For #169.

test cases:

```python
def test(a, c: int, w=4, *args: int, **kwargs: 'annotating kwargs') -> float:
    return 5.4

def test(a, b, c: int, w=4, *args: int, **kwargs: 'annotating kwargs') -> float:
    return 5.4

def test(c: int, w=4, *args: int, **kwargs: 'annotating kwargs') -> float:
    return 5.4

def test(a: float, c: int, *args: int, **kwargs: 'annotating kwargs') -> float:
    return 5.4

def test(a: float, c: int = 5, *args: int, **kwargs: 'annotating kwargs') -> float:
    return 5.4

def test(a: float, c: int, test=None) -> float:
    return 5.4

def test(*args: int, **kwargs) -> float:
    return 5.4

def test(x=55, *args: int, **kwargs) -> float:
    return 5.4

#def test(x=55, *args: int, y=5, **kwargs) -> float:
#    return 5.4
```

@rocky, I think we have an issue in parser in last commented test case. Could you please take a look?

Other test cases passed.


rocky:
Will check out the parse error. Thanks for doing! (And putting it a branch here to make it child-proof for me to pull).

PR #255, opened 2019-06-08
Add 3.6 STORE_ANNOTATION

x0ret:
Last 3.6 Opcode as discussed in #58 .


rocky:
@x0ret - I would like to put out a release of xdis and uncompyle6 tomorrow. The question I have you is whether to wait on this PR or merge it and continue (after release). Your thoughts?

x0ret:
@rocky - Thank you for your explanations which saves my time more and more.

> I would like to put out a release of xdis and uncompyle6 tomorrow. The question I have you is whether to wait on this PR or merge it and continue (after release). Your thoughts?

I think we can continue this PR after release, the PR usage is not urgent and it is just for 3.6 which i think barely used.

x0ret:
As you noted before, It was just a subscript rule. 
However in my testing i noticed there's a more indentation, For:

```python
class CV:
    var: ClassVar['CV']
```

but we got this:
```python
class CV:
        var: ClassVar['CV']
```

Since `ann_assign_init_value` and `ann_assign_not_init` have their corresponding indentation, I assumed we don't need `%|` in `store_annotation`.

rocky:
@x0ret Yes you are correct - that indentation should not have been there. Thanks for correcting in c871a4e.

PR #262, opened 2019-06-21
Reinstate some elif's

rocky:
@x0ret `n_ifelsestmtr` and `n_elifelsestmtr` need fixing, if you are interested. 

Also, since all of this is transformed by a subroutine,  duplicate table rules don't need to exist.

So: 

```
    'ifelsestmt':	( '%|if %c:\n%+%c%-%|else:\n%+%c%-', 0, 1, 3 ),
    'ifelsestmtc':	( '%|if %c:\n%+%c%-%|else:\n%+%c%-', 0, 1, 3 ),
    'ifelsestmtl':	( '%|if %c:\n%+%c%-%|else:\n%+%c%-', 0, 1, 3 ),
```

can become: 

```    
    'ifelsestmt':	( '%|if %c:\n%+%c%-%|else:\n%+%c%-', 0, 1, 3 ),
```

By setting `ifelsestmtl` to `ifelsestmt` in the code that changes the `kind` field. 

It is now clear to me how this got broken.  I changed some grammar rules to match the AST and to simplify and then the tests no longer triggered. 

This code was way too underdocumented for its trickiness (which I am not sure I fully understand in handling chained elif's). So I have started to document it better. 

Down the line though I would prefer this and the other tree-transformation processes in its own phase as is contemplated in decompile3.


x0ret:
Thanks. This was so helpful. 

I'm currently working on handling test cases like:

```python
def test():
    x = 3
    y = False
    if x < 2:
        pass
    elif x > 3:
        return x+2
    else:
        return x
```

in this branch.

x0ret:
@rocky, Could you please comment on this?

First of all we get this:

```
def test():
    x = 3
    y = False
    if x < 2:
        pass
    else:
        if x > 3:
            return x + 2
        return x
```

while i was trying to handle above example i observed `ifelsestmtr` grammer is not checked and matched in the parser phase instead it is parsed as `ifstmt + return` .

According to our grammer:

```
ifelsestmtr ::= testexpr return_if_stmts returns
```

I think it should be matched for the above example but we got this:

```
3. else_suite
                    suite_stmts (2)
                         0. stmt
                            ifstmt (2)
                                 0. testexpr
                                 1. _ifstmts_jump
                                    return_if_stmts (2)
                                         0. return_if_stmt (2)
                         1. stmt
                            return (3)
```

But it is exactly `ifelsestmtr` i suppose. After debugging i came to this grammer:

```
        sstmt ::= ifelsestmtr
```

and changing to :

```
        stmt ::= ifelsestmtr
```

fixes this issue and we have correct decompiled result:

```
def test():
    x = 3
    y = False
    if x < 2:
        pass
    elif x > 3:
        return x + 2
    else:
        return x
```

You thoughts?

rocky:
The short answer is that this is _probably_ the right thing to do. But I would test it out by running `test_pyenvlib --max 1000 --3.x.x --verify` with various values of 3.x.x. 

Here is the longer answer. (I probably spent way too much time than I have on this.) 

I really don't know why there is `sstmt` to begin with. You would have saved me some time indicating _which_ Python versions you are talking about. Apparently I can only find that rule in `parse3.py`. 

 - - - 

All of this indicates that down the line we need tests for decompilation to strings. The fact that there weren't any allowed me to make inadvertently break things when I changed the grammar.  (Since that word comes up a lot, grammar is spelled with an "A", rather than being something which is more "gramm" or a person or thing that "gram"s.)

x0ret:
Sorry, i forgot to add details.

This is [the line](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse3.py#L98) i was talking about.
I was working on Python 3.6.7 pyc.

rocky:
Ok - good to get confirmation. That is grammar rule is what I was looking at and I was trying 3.6ish too. 

But notice we don't have anything like this in Python 2.x which is odd because this is the version where this code originated. So this change will not change the behavior of 2.x and something additional may be needed there.

x0ret:
About Python 2.7 iit seems following diff works as expected:

```diff
..skipping...
diff --git a/uncompyle6/parsers/parse27.py b/uncompyle6/parsers/parse27.py
index b3228e35..9ce3a465 100644
--- a/uncompyle6/parsers/parse27.py
+++ b/uncompyle6/parsers/parse27.py
@@ -127,6 +127,8 @@ class Python27Parser(Python2Parser):

     def p_stmt27(self, args):
         """
+        stmt ::= ifelsestmtr
+
         # assert condition
         assert        ::= assert_expr jmp_true LOAD_ASSERT RAISE_VARARGS_1

@@ -178,6 +180,7 @@ class Python27Parser(Python2Parser):
         ifelsestmtc       ::= testexpr c_stmts_opt JUMP_ABSOLUTE else_suitec
         ifelsestmtl       ::= testexpr c_stmts_opt JUMP_BACK else_suitel
         ifelsestmtl       ::= testexpr c_stmts_opt CONTINUE else_suitel
+        ifelsestmtr ::= testexpr return_if_stmts COME_FROM returns

         # Common with 2.6
         return_if_lambda   ::= RETURN_END_IF_LAMBDA COME_FROM
```
And we get this:
```python
def test():
    x = 2
    y = False
    if x < 2:
        pass
    elif x > 3:
        return x + 2
    else:
        return x
# okay decompiling ../tests
```
Sorry about the rules, they may not be too accurate.
I won't check out other versions though. Will check later.

rocky:
Thanks again for following through and fixing.  

Ultimately whether this should be in `parse27.py` or earlier I guess depends on whether this is needed or relevant for earlier versions. Versions of this that handled Python up to 2.4 didn't require this. For 2.5 and 2.6 it is not clear - some versions have this but it is not clear if those were needed or well tested. It may be that this is just because it was based on 2.7 code. 

So I think the overall plan here is to try to fix this here, and tidy it up a little bit. Code like:

```
        if len(return_stmts_node[0]) >= 3:
            if (return_stmts_node[0][-1][0] == 'ifstmt'
                and return_stmts_node[0][-1][0][1][0] == 'return_if_stmts'):
                if_ret_at_end = True
```

is pretty indecipherable. 

Then with this working, I'd like to see it redone in decompile3 using this code as a guide.

x0ret:
Python 2.6(.9) does not have that `COME_FROM` and needs this:

```diff
@@ -102,6 +102,8 @@ class Python26Parser(Python2Parser):

     def p_stmt26(self, args):
         """
+        stmt ::= ifelsestmtr
+
         # We use filler as a placeholder to keep nonterminal positions
         # the same across different grammars so that the same semantic actions
         # can be used
@@ -173,6 +175,9 @@ class Python26Parser(Python2Parser):
         iflaststmt     ::= testexpr_then c_stmts_opt JUMP_ABSOLUTE come_froms POP_TOP
         iflaststmt     ::= testexpr      c_stmts_opt JUMP_ABSOLUTE come_froms POP_TOP

+        ifelsestmtr       ::= testexpr_then return_if_stmts returns
+
         testexpr_then  ::= testtrue_then
         testexpr_then  ::= testfalse_then
         testtrue_then  ::= expr jmp_true_then
```

About Python 2.5(.6) without any changes it produces correct decompilation but in another form!

> is pretty indecipherable.

I will check that function and try to simplify and document things to clarify more later.

x0ret:
Sorry for the delay, I've been busy lately.

@rocky, i'm confused about one thing: `ifelsestmtr` is telling us i have a `return` as last stmt or i have a `return` you should handle yourself? 
According to these lines:

```python
        if len(return_stmts_node[0]) >= 3:
            if (return_stmts_node[0][-1][0] == 'ifstmt'
                and return_stmts_node[0][-1][0][1][0] == 'return_if_stmts'):
                if_ret_at_end = True
```

I think the latter seems to be correct though. I need this clarification to debug correctly. 

In my testings i noticed there are some cases we are not handling correctly.

Case#1: 
```python
def test():
    x = 3
    y = False
    if x < 2:
        pass
    elif x > 3 :
        return x+2
    else:
        print(x)
```

Case#2:

```python 
def test():
    x = 3
    y = False
    if x < 2:
        return x-1
    elif x > 3 :
        pass
    else:
        print(x)
```

Which i think i can handle afterward. Thanks

rocky:
> Sorry for the delay, I've been busy lately.

delay and busy happen. Happens to me all the time. No need to apologize. 

> `ifelsestmtr` is telling us i have a return as last stmt or i have a return you should handle yourself?

I don't understand your question

From `parse.py`: 
```
        # "returns" nonterminal is a sequence of statements that ends in a RETURN statement.
```

As for what

```python
if len(return_stmts_node[0]) >= 3:
            if (return_stmts_node[0][-1][0] == 'ifstmt'
                and return_stmts_node[0][-1][0][1][0] == 'return_if_stmts'):
                if_ret_at_end = True
```
does and whether it is correct, I don't know. `return_if_stmts` is a nonterminal that I believe ends in a `return` statement too. It may be that at one time this was correct but not anymore. 

As far as Cases #1 & #2 just handle what can and we can come back to these.

x0ret:
I need more work on this to fix both cases, i will handle these cases later.

About the failed test, I'm trying to fix.

x0ret:
@rocky, could you please check `pytest/test_grammar.py` change?

rocky:
> @rocky, could you please check `pytest/test_grammar.py` change?

It looks to me like something is wrong in `pytest/test_grammar.py` because `ifelsestmtr` appears both on the left- and right-hand sides of rules in Python 3. 

The test is there to warn of inadvertent changes and since we've looked this over and things look okay, I am okay with doing things like what you did, but I'd make a note to come back and review the test for why the checker needs that. It looks like in later tests though you need to narrow when to add this, which is also weird.

rocky:
I looked at Case#1. The reason that isn't caught is because we are not decompiling it correctly to begin with. It is coming out: 

```python
    x = 3
    y = False
    if x < 2:
        pass
    else:
        if x > 3:
            return x + 2
        print(x)
        # ^^^^ wrong indent.

```

rocky:
As for Case#2 (on 3.7.3 as I did with Case#1) we are translating that into something that is in my opinion just as good as using `elif` to the extent that there are not extra indents. 

```
def test():
    x = 3
    y = False
    if x < 2:
        return x - 1
    if x > 3:
        pass
    else:
        print(x)
```

In theory additional transformations could be done to tun this into `elif` but I don't think it needed here.

x0ret:
> It looks to me like something is wrong in pytest/test_grammar.py because ifelsestmtr appears both on the left- and right-hand sides of rules in Python 3.

Sorry, my bad. `stmt ::= ifelsestmtr` was missing in `parse27.py`.

Thanks for your time.

rocky:
> Sorry, my bad. `stmt ::= ifelsestmtr` was missing in parse27.py.

I am always amazed at how tricky this stuff is. I didn't catch that. 

But this emphasizes the usefulness for tests that check that decompilng some bytecode produces a source code string. It could be done in pytest as we do to check disassembly.

> Thanks for your time.

Actually, I thank you for all of the work spent on this. The project has moved along farther in the last 2-3 week than it has in the last 6 months. I am optimistic that we will get to the point where most of the known problems are addressed. 

With respect to this, if we want to commit now, go ahead. If instead you want to add more to this, that's okay too. Up to you.

x0ret:
Thank you again.

> With respect to this, if we want to commit now, go ahead. If instead you want to add more to this, that's okay too. Up to you.

I think i'm not ready to add more to this, So i can come back to these later. If you are okay with it, please merge this, i added cases to my TODOs.

PR #263, opened 2019-06-23
WIP pypy3.6 handling

rocky:
Some of the semantic-action phase calls don't work, but otherwise I think okay.

And yes, this adds more mess. A cleanup would be nice.


rocky:
There is still a little more to do, but that will be done in the master branch.

PR #264, opened 2019-06-24
if/elif for 2.5-2.7

rocky:
Specifically simple_source/03_if_elif.py


x0ret:
I totally missed it. Thanks.

Issue #228, opened 2019-05-09
Inconsistent while block due to grammar? in python3


x0ret:
## Description

Suppose following src:
```python
def func():
    while not x:
        print(x)
```

`uncompyle6`  decompiles this to :

```
def func():
    while 1:
        if not x:
            print(x)
```

I think we need a better representation or at least a `break` statement.

based on my limited debugging experience here, Please consider this module disassembly:

```
   3       0  SETUP_LOOP           18  'to 18'
           2  LOAD_GLOBAL              'x'
           4  POP_JUMP_IF_TRUE     16  'to 16'

   4       6  LOAD_GLOBAL              'print'
           8  LOAD_GLOBAL              'x'
          10  CALL_FUNCTION         1  ''
          12  POP_TOP
          14  JUMP_BACK             2  'to 2'
          16  POP_BLOCK
        18_0  COME_FROM_LOOP        0  '0'
          18  LOAD_CONST               None
          20  RETURN_VALUE
```

And we have following grammer in `parse3.py` at [L#154](https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse3.py#L154):

```
        iflaststmt ::= testexpr c_stmts_opt JUMP_ABSOLUTE

        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK [1]
        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK COME_FROM_LOOP
        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK POP_BLOCK [2]
```

based on disassembly if we omit [1], [2] (commenting them out) we have following code:

```
def func():
    while not x:
        print(x)
```

I'm not sure what is the best fix here. I also checked 3.7 version result is the same.

for reference we have following grammer before fix(?):

```
         2-12  iflaststmtl ::= testexpr c_stmts_opt (7)
L.  3:   6     _ifstmts_jump ::= c_stmts_opt (7)
         2     lastl_stmt ::= iflaststmtl (7)
         2-12  ifstmt ::= testexpr _ifstmts_jump (7)
         2     l_stmts ::= lastl_stmt (7)
         2     stmt ::= ifstmt (7)
         2     _stmts ::= stmt (7)
         2-14  iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK (8)
         2     lastl_stmt ::= iflaststmtl (8)
         2     l_stmts ::= lastl_stmt (8)
[...]

PR #265, opened 2019-06-25
[WIP] Handling if elif else more

rocky:
@x0ret 

Here is yet another transformation for handing `if .. `elif .. `else` which is more involved that occurs in Python 3.5 for this:

```python
x = 1
while 1:
    if x != 2:
        if __file__:
            x = 2
        elif __name__:
            pass
        else:
            break

while x != 2:
    if __file__:
        x = 2
    elif __name__:
        pass
    else:
        break
```

In a separate issue, Python 3.5 gets the `while 1` vs.`while x != 2` backwards, whiile python 3.4 turns them both into `while 1` and misses one of the the two `elif` opportunities.

### DIFF
--- a/uncompyle6/semantics/pysource.py
+++ b/uncompyle6/semantics/pysource.py
@@ -674,7 +674,7 @@ class SourceWalker(GenericASTTraversal, object):
             n = n[0][0]
         elif n[0].kind in ('lastc_stmt', 'lastl_stmt'):
             n = n[0]
-            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl'):
+            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl', 'ifelsestmtc'):
                 # This seems needed for Python 2.5-2.7
                 n = n[0]
                 pass
@@ -682,7 +682,7 @@ class SourceWalker(GenericASTTraversal, object):
         elif ( len(n) > 1 and 1 == len(n[0]) and n[0] == 'stmt'
                and n[1].kind == "stmt" ):
             else_suite_stmts = n[0]
-            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if else_suite_stmts[0].kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
                 if not preprocess:
                     self.default(node)
                 return
--- a/uncompyle6/semantics/pysource.py
+++ b/uncompyle6/semantics/pysource.py
@@ -661,21 +661,33 @@ class SourceWalker(GenericASTTraversal, object):
           if ..
           elif ...
 
+          [else ...]
+
         where appropriate
         """
         else_suite = node[3]
 
         n = else_suite[0]
+        old_stmts = None
 
         if len(n) == 1 == len(n[0]) and n[0] == 'stmt':
             n = n[0][0]
         elif n[0].kind in ('lastc_stmt', 'lastl_stmt'):
             n = n[0]
-            if n[0].kind in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl'):
                 # This seems needed for Python 2.5-2.7
                 n = n[0]
                 pass
             pass
+        elif ( len(n) > 1 and 1 == len(n[0]) and n[0] == 'stmt'
+               and n[1].kind == "stmt" ):
+            else_suite_stmts = n[0]
+            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+                if not preprocess:
+                    self.default(node)
+                return
+            old_stmts = n
+            n = else_suite_stmts[0]
         else:
             if not preprocess:
                 self.default(node)
@@ -695,6 +707,18 @@ class SourceWalker(GenericASTTraversal, object):
             elif n.kind in ('ifelsestmt', 'ifelsestmtc', 'ifelsestmtl'):
                 n.kind = 'elifelsestmt'
         if not preprocess:
+            if old_stmts:
+                if n.kind == "elifstmt":
+                    trailing_else = SyntaxTree("stmts", old_stmts[1:])
+                    # We use elifelsestmtr because it has 3 nodes
+                    elifelse_stmt = SyntaxTree(
+                        'elifelsestmtr', [n[0], n[1], trailing_else])
+                    node[3] = elifelse_stmt
+                    pass
+                else:
+                    # Other cases for n.kind may happen here
+                    return
+                pass
             self.default(node)
 
     n_ifelsestmtc = n_ifelsestmtl = n_ifelsestmt


RamazanKurnaz:
tenkyu bro :)

x0ret:
Thanks, @rocky i think this diff manages the two cases for 3.5:

```python
diff --git a/uncompyle6/semantics/pysource.py b/uncompyle6/semantics/pysource.py
index ce540409..c3cbd972 100644
--- a/uncompyle6/semantics/pysource.py
+++ b/uncompyle6/semantics/pysource.py
@@ -674,7 +674,7 @@ class SourceWalker(GenericASTTraversal, object):
             n = n[0][0]
         elif n[0].kind in ('lastc_stmt', 'lastl_stmt'):
             n = n[0]
-            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl'):
+            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl', 'ifelsestmtc'):
                 # This seems needed for Python 2.5-2.7
                 n = n[0]
                 pass
@@ -682,7 +682,7 @@ class SourceWalker(GenericASTTraversal, object):
         elif ( len(n) > 1 and 1 == len(n[0]) and n[0] == 'stmt'
                and n[1].kind == "stmt" ):
             else_suite_stmts = n[0]
-            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if else_suite_stmts[0].kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
                 if not preprocess:
                     self.default(node)
                 return
```

Could you please verify?

About the 3.4, i'm working on it but it's not good, will try later.

rocky:
Hmm....

This: 
```
-            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if else_suite_stmts[0].kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
```

would mean the previous change got something wrong. I retested without the change and don't see the problem. What exactly is `else_suite_stmts`? And what python version(s) was this for? 

Also both of these fail on: 

```python
x = 1
while 1:
    if x != 2:
        if __file__:
            x = 2
        elif __name__:
            pass
        else:
            x = 5
            break

while x != 2:
    if __file__:
        x = 2
    elif __name__:
        pass
    else:
        x = 5
        break
```

x0ret:
I was working on 3.5, 3.6, for e.g. on 3.5 for the first one i got this:

```python
x = 1
while x != 2:
    if __file__:
        x = 2
    else:
        if __name__:
            continue
        x = 5
        break
```

however with the change i get this:

```python
x = 1
while x != 2:
    if __file__:
        x = 2
    elif __name__:
        continue
    else:
        x = 5
        break
```

And the `else_suite_stmts` is :

```
                             3. else_suitel
                                l_stmts (3)
                                     0. stmt
                                        ifstmt (2)
                                             0. testexpr
                                                testfalse (2)
                                                     0. expr
                                                        L.   6      36  LOAD_NAME                __name__
                                                     1. jmp_false
                                                                    39  POP_JUMP_IF_FALSE    45  'to 45'
                                             1. _ifstmts_jump (2)
                                                 0. c_stmts
                                                    stmt
                                                        continue
                                                            L.   7      42  CONTINUE              9  'to 9'
                                                 1.             45  ELSE                     '52'
                                     1. stmt
                                        assign (2)
                                             0. expr
                                                L.   9      45  LOAD_CONST               5
                                             1. store
                                                            48  STORE_NAME               x
                                     2. stmt
                                        break
                                            L.  10      51  BREAK_LOOP
```

rocky:
@x0ret  Ok - yes I see this now. This is great.  The `pass` turns into `continue`, but that's fine.  And the example I gave above works too!

rocky:
I must have applied the patch incorrectly?  So please add it to this branch.

x0ret:
@rocky, I think we have an issues with having these three rules (`parse3.py`) and i know we discussed about it in #228 : 

```
        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK
        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK COME_FROM_LOOP
        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK POP_BLOCK
```

Please let me explain the case in detail, As for 3.4 (And also for others) suppose following example borrowed from above test cases:

```python
while x != 2:
    if __file__:
        x = 2
    elif __name__:
        pass
    else:
        x = 5
        break
```

and we get this:

```python
while 1:
    if x != 2:
        if __file__:
            x = 2
    elif __name__:
        continue
        x = 5
        break
```

however let me follow disassembly and tree a bit:

```
                             3. else_suitel
                                l_stmts
                                    lastl_stmt
                                        iflaststmtl (4)
                                             0. testexpr
                                                testfalse (2)
                                                     0. expr
                                                        L.  15      89  LOAD_NAME                __name__
                                                     1. jmp_false
                                                                    92  POP_JUMP_IF_FALSE    98  'to 98'
                                             1. c_stmts (3)
                                                 0. stmt
                                                    continue
                                                        L.  16      95  CONTINUE             62  'to 62'
                                                 1. stmt
                                                    assign (2)
                                                         0. expr
                                                            L.  20      98  LOAD_CONST               5
                                                         1. store
                                                                       101  STORE_NAME               x
                                                 2. stmt
                                                    break
                                                        L.  21     104  BREAK_LOOP
                                             2.            105  JUMP_BACK            62  'to 62'
                                             3.            108  POP_BLOCK
                 2.          109_0  COME_FROM_LOOP       59  '59'
```

I think the `JUMP_BACK` and `POP_BLOCK` should not be seen in the `iflaststmtl`, instead as of offsets they are belongs to `while`. But why i'm saying this?

1. instead of `while x != 2` we have `while 1: ...` which can be avoided, since according to following rule this can be achieved by removing `iflaststmtl` rules which can be converted to `ifelsestmtl`:

```
        whilestmt         ::= SETUP_LOOP testexpr l_stmts_opt JUMP_BACK POP_BLOCK
                              COME_FROM_LOOP
```
2. As you stated about the fact that it converts `pass` to `continue`, i traced the code and came to this rule that can be workaround the issue (Should be added):

```
ifelsestmtl ::= testexpr c_stmts_opt continue else_suitel
```

So after using above changes we get following decompilation for 3.4:

```
while x != 2:
    if __file__:
        x = 2
    elif __name__:
        pass
    else:
        x = 5
        break
```

Anyway you stated before that it could break other cases. But i am ready to check all test cases and trace if the above solution works (Since we can fix #228, and also have a better decompilation for `while` even for other cases).

However At the end we can workaround this by patching conditionally for `iflaststmtl` but i hope we avoid such a behavior:

```diff
        if n.kind in ('ifstmt', 'iflaststmt', 'iflaststmtl'):
            node.kind = 'ifelifstmt'
            n.kind = 'elifstmt'
+            if n[1].kind == 'c_stmts' and n[1][0].kind == 'stmt' and n[1][0][0].kind == 'continue':
+                old_stmts = n[1]
+                n[1] = SyntaxTree("c_stmts", [n[1][0]])
``` 
What do you think?

Sorry for this long comment. Thanks for reading.

x0ret:
Using this diff i passed `make check`:

```diff
diff --git a/uncompyle6/parsers/parse3.py b/uncompyle6/parsers/parse3.py
index 34a04bf7..8bee4320 100644
--- a/uncompyle6/parsers/parse3.py
+++ b/uncompyle6/parsers/parse3.py
@@ -154,9 +154,9 @@ class Python3Parser(PythonParser):

         iflaststmt ::= testexpr c_stmts_opt JUMP_ABSOLUTE

-        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK
-        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK COME_FROM_LOOP
-        iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK POP_BLOCK
+        # iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK
+        # iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK COME_FROM_LOOP
+        # iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK POP_BLOCK

         # These are used to keep parse tree indices the same
         jump_forward_else  ::= JUMP_FORWARD ELSE
@@ -182,6 +182,8 @@ class Python3Parser(PythonParser):

         ifelsestmtl ::= testexpr c_stmts_opt JUMP_BACK else_suitel
         ifelsestmtl ::= testexpr c_stmts_opt cf_jump_back else_suitel
+        ifelsestmtl ::= testexpr c_stmts_opt continue else_suitel
+

         cf_jump_back ::= COME_FROM JUMP_BACK

@@ -348,6 +350,8 @@ class Python3Parser(PythonParser):

     def p_loop_stmt3(self, args):
         """
+        stmt ::= whileelsestmt2
+
         for               ::= SETUP_LOOP expr for_iter store for_block POP_BLOCK
                               COME_FROM_LOOP

@@ -363,6 +367,8 @@ class Python3Parser(PythonParser):
         whilestmt         ::= SETUP_LOOP testexpr l_stmts_opt COME_FROM JUMP_BACK POP_BLOCK
                               COME_FROM_LOOP

+        whilestmt         ::= SETUP_LOOP testexpr l_stmts_opt JUMP_BACK POP_BLOCK JUMP_BACK
+                              COME_FROM_LOOP
         whilestmt         ::= SETUP_LOOP testexpr l_stmts_opt JUMP_BACK POP_BLOCK
                               COME_FROM_LOOP

@@ -375,6 +381,9 @@ class Python3Parser(PythonParser):
         whileelsestmt     ::= SETUP_LOOP testexpr l_stmts_opt JUMP_BACK POP_BLOCK
                               else_suitel COME_FROM_LOOP

+        whileelsestmt2     ::= SETUP_LOOP testexpr l_stmts_opt  JUMP_BACK POP_BLOCK
+                              else_suitel JUMP_BACK COME_FROM_LOOP
+
         whileTruestmt     ::= SETUP_LOOP l_stmts_opt          JUMP_BACK POP_BLOCK
                               COME_FROM_LOOP

diff --git a/uncompyle6/semantics/consts.py b/uncompyle6/semantics/consts.py
index cae5c952..214b4e8f 100644
--- a/uncompyle6/semantics/consts.py
+++ b/uncompyle6/semantics/consts.py
@@ -377,6 +377,7 @@ TABLE_DIRECT = {
     'while1stmt':	    ( '%|while 1:\n%+%c%-\n\n', 1 ),
     'while1elsestmt':   ( '%|while 1:\n%+%c%-%|else:\n%+%c%-\n\n', 1, -2 ),
     'whileelsestmt':	( '%|while %c:\n%+%c%-%|else:\n%+%c%-\n\n', 1, 2, -2 ),
+    'whileelsestmt2':	( '%|while %c:\n%+%c%-%|else:\n%+%c%-\n\n', 1, 2, -3 ),
     'whileelselaststmt':	( '%|while %c:\n%+%c%-%|else:\n%+%c%-', 1, 2, -2 ),

     # Note: Python 3.8+ changes this
diff --git a/uncompyle6/semantics/pysource.py b/uncompyle6/semantics/pysource.py
index ce540409..aeed7113 100644
--- a/uncompyle6/semantics/pysource.py
+++ b/uncompyle6/semantics/pysource.py
@@ -674,7 +674,7 @@ class SourceWalker(GenericASTTraversal, object):
             n = n[0][0]
         elif n[0].kind in ('lastc_stmt', 'lastl_stmt'):
             n = n[0]
-            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl'):
+            if n[0].kind in ('ifstmt', 'iflaststmt', 'iflaststmtl', 'ifelsestmtl', 'ifelsestmtc'):
                 # This seems needed for Python 2.5-2.7
                 n = n[0]
                 pass
@@ -682,7 +682,7 @@ class SourceWalker(GenericASTTraversal, object):
         elif ( len(n) > 1 and 1 == len(n[0]) and n[0] == 'stmt'
                and n[1].kind == "stmt" ):
             else_suite_stmts = n[0]
-            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if else_suite_stmts[0].kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
                 if not preprocess:
                     self.default(node)
                 return
```

rocky:
Thanks for the detailed analysis. It is going to take me some time to think about this, review the history, make and some tests. 

Right now I am swamped with other work. 

We will eventually get to the bottom of this though....

x0ret:
Thanks

> We will eventually get to the bottom of this though....

I pushed the changes for 3.5. So we can continue and discuss about `while` case later.

rocky:
I need to do a lot more work to investigate more, but here is what I understand so far. 

In Python 3.4 the difference between that particular  `while 1` and `while` is that there is an extra backward `JUMP_ABSOLUTE` instruction (shown as `JUMP_BACK`) in the `while 1`.  One (but only one) of the `then` clauses jumps to the second or looping one while the first seems to be part of dead code after the `BREAK_LOOP`.

I am guessing in later Pythons the dead code and/or jumps to jumps is removed. 

But going on the principle of trying to model the code more honestly and accurately in the grammar (and we saw how this is already botched with the `COME_FROM`s that come from the wrong place, it would be good, if possible, to keep 

```
iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK
```

if that works. If other  `JUMP_BACK`s need to get removed in places they don't belong that would be good too, but right now I am not sure if that is the situation. 

The other general thought that comes to mind is that this should probably be localized into the versions where this occurs. Specifically 3.4 and removed in 3.5 if that no longer is relevant because code generation and jump optimization has changed. 

But again I need to study what's up in more detail and make sure that in the end we capture the semantics context of the instructions better inside the tree.

rocky:
```
iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK COME_FROM_LOOP
```

is most pretty definitely wrong since the `COME_FROM_LOOP` and the ` JUMP_BACK` should be part of the loop, not part of any `if` statement, whether or not it has ends in  `JUMP_BACK` or not. 

So commenting out or removing that, is probably correct.

x0ret:
> is most pretty definitely wrong since the COME_FROM_LOOP and the JUMP_BACK should be part of the loop, not part of any if statement, whether or not it has ends in JUMP_BACK or not.

I was reading Python `compile.c` and according to `compiler_while`:

```c
    ADDOP_JREL(c, SETUP_LOOP, end);
    compiler_use_next_block(c, loop);
...
    if (constant == -1) {
        VISIT(c, expr, s->v.While.test);
        ADDOP_JABS(c, POP_JUMP_IF_FALSE, anchor);
    }
    VISIT_SEQ(c, stmt, s->v.While.body);
    ADDOP_JABS(c, JUMP_ABSOLUTE, loop);
...
    if (constant == -1)
        compiler_use_next_block(c, anchor);
    ADDOP(c, POP_BLOCK);
...
    if (orelse != NULL) /* what if orelse is just pass? */
        VISIT_SEQ(c, stmt, s->v.While.orelse);
    compiler_use_next_block(c, end);
```

for a *NOT-WHILE-0* case we have following order:
```
SETUP_LOOP    testexpr   POP_JUMP_IF_FALSE    stmt   JUMP_ABSOLUTE  POP_BLOCK [elsestmt]
```
and with using `COME_FROM_LOOP` as end `bb` for the end of the rule. So having variant of while (end of the stmt bytecodes) i think we have conflict with `iflaststmtl` as `l_stmts` which results in 3.4 situation.

However one question comes to my mind, Here we are trying to match exactly the Python AST and some analysis like above or to have an correct grammer as much as possible?

** Python `compile.c` borrowed from 3.4 tag and 3.6 tag (both identical) **

rocky:
Thanks again for looking more deeply into this. Excellent Idea to look at the source code that generates this! I am hitting myself in the head for not thinking of that before. 

I don't understand what is meant by: 

> i think we have conflict with iflaststmtl as l_stmts which results in 3.4 situation.

If you mean that there are two grammar rules that can match, then that is not a "conflict". It is just a situation we have to deal with. This is one area where this is different than conventional {LA,S,}LR, LL parsing. Recall we are using a  J Earley Algorithm parser which allows a sequence of tokens to be parsed several ways.  That's like translating from one language to another and having two different translations. 

The question is however whether one of the translations is _wrong_. Is it? And if so is it because of the pattern or because of something else.  For example: `LOAD_CONST expr`  might be parsed as a named parameter and value. But if the value of the first `LOAD_CONST` is an integer then even though there is a match, the problem isn't in the grammar. We recently changed this in many places to `LOAD_STR expr` which would remove that first possibility. However if the `LOAD_STR` has value `"I feel pretty"` then this isn't a valid keyword parameter, even though the more restrictive rule again matches. 

So to come back to the problem at hand is the rule for `iflaststmtl` _wrong_ in some way? 

As you observed, the `COME_FROM_LOOP` is just another way of indicating what is shown in disassembly as `>>` &mdash; that this is is a target of a jump somewhere. Recall that in the first `JUMP_ABSOLUTE`,  or the one that should be part of `iflaststmtl`, that _wasn't_ the target of a jump.  

I believe that the `JUMP_ABSOLUTE` in `iflaststmtl` should _never_ the target of a loop. That instead it is a fall-though jump back to the loop &mdash; and that is why the rule specifies only matching `JUMP_BACK`. And there should never be a `COME_FROM` before that. 

And finally to your question:

> Here we are trying to match exactly the Python AST and some analysis like above or to have an correct gramm[a]r as much as possible?

First of all this isn't an AST in the sense that some of the tokens are stripped away like set braces around set literals are Python's AST or ASTs in general &mdash; _every instruction has a parent in our (initial) syntax tree_. 

But then you might ask what about the (implicit) tree that seems to be implied by `compile.c`?  Again I don't think that is possible, except in the most extreme of cases where there is no code optimization. Once you have things replacing jumps to jumps with a single jump (as I believe happens here somewhere down the line in the more recent Pythons), then all bets are off. 

That said, using how Python generates code, and using the declared grammar are all good things to guide us. And using AST prefixes  (`SetComp` -> `set_comp`) are good. But as you saw we may need several kinds of `set_comp` where the AST is fine with one version where maybe there is an attribute in that node that distinguishes different kinds if need be. 

> _Anyway this is a brilliant journey, Thanks for letting me in :-) _

I am glad you are enjoying, And the code is all the better for your joining.

rocky:
I'm going though this again. 

```
iflaststmtl ::= testexpr c_stmts_opt JUMP_BACK
```

is used in `3.3.7/lib/python3.3/difflib.py` specifically, line 977:

```python
            for i in range(alo, ahi):
                ai = a[i]
                if ai == bj:
                    if eqi is None:
                        eqi, eqj = i, j
                    continue

``` 

By the way, please, instead of posting diffs, make a new branch and off of this one. Otherwise I have to cut and paste the diff, and in addition to the additional work, there is a chance that I'll get it wrong somehow. Thanks. 

Right now I am having trouble getting this: 

```diff
diff --git a/uncompyle6/semantics/pysource.py b/uncompyle6/semantics/pysource.py
index ce540409..aeed7113 100644
@@ -682,7 +682,7 @@ class SourceWalker(GenericASTTraversal, object):
         elif ( len(n) > 1 and 1 == len(n[0]) and n[0] == 'stmt'
                and n[1].kind == "stmt" ):
             else_suite_stmts = n[0]
-            if else_suite_stmts.kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
+            if else_suite_stmts[0].kind not in ('ifstmt', 'iflaststmt', 'ifelsestmtl'):
                 if not preprocess:
                     self.default(node)
                 return
```
 
to apply, so in my next branch on this branch, I'll leave it out.

x0ret:
> By the way, please, instead of posting diffs, make a new branch and off of this one. Otherwise I have to cut and paste the diff, and in addition to the additional work, there is a chance that I'll get it wrong somehow. Thanks.

Sure, so sorry about that.

Anyway i had commited that change in this branch at `203d118` but forgot to remove it from the diff above.

rocky:
Sorry for the mess. Let's work in the other PR #269 and I guess close this?

x0ret:
Sure, Thanks.

PR #266, opened 2019-06-25
[WIP] convert the rest of the LOAD_CONST->LOAD_STR

rocky:
@x0ret FYI Not totally sure it is worth the small bit of elegance and potential breakage.


rocky:
This just isn't working out.

PR #276, opened 2019-07-10
Fixes #274

### LINKED ISSUE #274: Chained comparisons in `if` or `and`
In Python 3.7.3 

```
digit = 1
if isinstance(digit, int) and 0 <= digit <= 9:
    pass
```
or

```
def chained_compare_b(a, obj):
    if a:
        if -0x80000000 <= obj <= 0x7fffffff:
            return 5

chained_compare_b(True, 0x0)
```

are not coming out correctly

x0ret:
Fixes #274 

It passed all the test cases, Please let me know if there should be changes.

Thanks.

### DIFF
--- a/uncompyle6/parsers/parse37.py
+++ b/uncompyle6/parsers/parse37.py
@@ -130,7 +130,7 @@ class Python37Parser(Python36Parser):
         jmp_false37                ::= POP_JUMP_IF_FALSE COME_FROM
         list_if                    ::= expr jmp_false37 list_iter
 
-        _ifstmts_jump              ::= c_stmts_opt come_froms
+        ifstmt                     ::= testexpr _ifstmts_jump COME_FROM
 
         and_not                    ::= expr jmp_false expr POP_JUMP_IF_TRUE
 


x0ret:
Sorry about that i always forget about `verify-run`. I'm handling this, thanks.

x0ret:
For now we are passing `make check-bytecode-3.7` with this change, however i'm not sure it's the best choice and i'd be grateful if you comment on this in your free time.

Anyway having `come_froms` in `_ifstmt_jump` seems to force nested if (parent `ifstmt`'s) statement to be `pass` stmt which we are trying to avoid in this PR.

Thanks

x0ret:
I get multiple parse error. I'll take a look into this more carefully. Thanks.

x0ret:
I get multiple parse error. I'll take a look into this more carefully. Thanks.

x0ret:
@rocky, i failed to pass 3.7 library tests even without this changes and eventually i recall the #235 issue which i failed to complete. 
So i think we can't count on these tests unless we fixed the 3.7 library full decompilation.

What do you think?

rocky:
> So i think we can't count on these tests unless we fixed the 3.7 library full decompilation.
>
> What do you think?

Yeah, I was aware of incompletness starting at around 3.5 or 3.6 or so. So I was interested, overall, whether the count got larger or smaller. That is, whether by doing this we have just traded off some bugs for some other bugs. That is in fact how in the past I have decided whether I should put in a change like this when there is no clear explanation from first principles as to why this change is better. 

For example, in a recent change I think you proposed, I could see from the grammar that a `COME_FROM` token after a grammar rule was clearly wrong, and I gave a reason why I thought that was the case. 

If we you could show why the existing rules that get removed are wrong, and how the new rules are better, then of course that is a stronger case than the statistical one I was asking about. I am not saying that is not the situation here, I don't know and haven't done the work to verify.

In the broader sense though as I have written elsewhere, better would be to attack what is really the main problem head on. It is going to be a lot of work, and it is hard in that sense. But if we don't continue, we'll never finish it.

rocky:
I am optimistic that we'll find a better way to handle this. Closing this since it fixes some things and possibly breaks others.

PR #279, opened 2019-07-13
Fix issue in BUILD_LIST_UNPACK and variants

x0ret:
Also handle star tuples to unpack correctly as we've done in 3.6 `format_pos_args`.


x0ret:
@rocky, I used following test cases, Could you please verify the fix?

```python
a = 5
x = [1, 2, 3]
i = [(a,), x]
j = [a, *x]

def f(x):
    return [1, 2, *x, y, *z, *v, *u, *q]
```

uncompyle6/semantics/pysource.py
```diff
@@ -1822,7 +1822,7 @@ def n_list(self, node):
         # will assume that if the text ends in *.
         last_was_star = self.f.getvalue().endswith("*")
 
-        if lastnodetype.endswith("UNPACK"):
+        if lastnodetype.startswith("BUILD_LIST_UNPACK"):
             # FIXME: need to handle range of BUILD_LIST_UNPACK
```
rocky: Probably can get rid of FIXME now that you have fixed this :-)

uncompyle6/semantics/pysource.py
```diff
@@ -1822,7 +1822,7 @@ def n_list(self, node):
         # will assume that if the text ends in *.
         last_was_star = self.f.getvalue().endswith("*")
 
-        if lastnodetype.endswith("UNPACK"):
+        if lastnodetype.startswith("BUILD_LIST_UNPACK"):
```
rocky: This is more an observation rather than something necessarily something to change: BUILD_LIST_UNPACK was added in 3.5.

x0ret:
Thanks for all your suggestions. I tried to cover all of them. Please let me know if there are changes that shouldn't be.

rocky:
Thanks! Let's do this!

PR #282, opened 2019-07-30
Tree transforms for assert and elif

rocky:
Adds a tree-transformation phase where we can simplify the AST such as turning an 
`if ... raise` into an `assert`  and `if ..  else if ..`  into `if .. elif ... `. 

See corresponding branch in decompyle3.


rocky:
This is a mess. Will revisit some other time.

PR #286, opened 2019-08-21
Add tree transformation phase...

rocky:
if ... else if ... ->  if ... elif ..
if .. : raise AsssertionError ->  assert

Add options --tree=before --tree=after  -T expanded to include this

This code ported from decompyle3. x0ret did all the heavy lifting.


rocky:
@x0ret - I couldn't get the docstring stuff from decompyle3 to port over. I was getting a failure in the docstring test. If you could port it over I'd be grateful. Otherwise we can wait on that.

PR #339, opened 2020-12-28
parsers: parse2: fix: also emit expr32 if count perfectly divisible by 1024

bloerwald:
`expr1024` requires `expr32`, but a `build_count` of 1024 would emit only the
`expr1024` rule and rely on luck of it being emitted somewhere else.

Emit `expr32` rule either if there is a `expr32` use or a `expr1024` use to avoid.

### DIFF
--- a/uncompyle6/parsers/parse2.py
+++ b/uncompyle6/parsers/parse2.py
@@ -319,7 +319,7 @@ class Python2Parser(PythonParser):
                 build_count = token.attr
                 thousands = build_count // 1024
                 thirty32s = (build_count // 32) % 32
-                if thirty32s > 0:
+                if thirty32s > 0 or thousands > 0:
                     rule = "expr32 ::=%s" % (" expr" * 32)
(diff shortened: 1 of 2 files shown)


bloerwald:
Note: I haven't used the python3 version, so it may be required to also detect the cases in https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse36.py#L285, https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse37.py#L1278, https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/parse37base.py#L403, which also unconditionally rely on `expr32`/`expr1024` being present. This may already be implicitly checked though, I didn't have a closer look.

rocky (approved):
Looks right to me. Thanks!

rocky:
I'll also throw out the idea that if you are running into this bug, you are probably deparsing some game data. (But I won't tell anyone.)

PR #348, opened 2021-06-14
'NoneType' object is not iterable with numproc > 1

IzeBerg:
main calls with source_paths=None, but it needs to be iterable


rocky:
Thanks!

PR #365, opened 2021-11-02
Convert 14 statements to the usage of augmented assignments

elfring:
[Augmented assignment statements became available with Python 2](https://docs.python.org/3/whatsnew/2.0.html#augmented-assignment "Augmented assignments").
:thought_balloon: Thus I suggest to adjust 14 source code places accordingly.

### DIFF
--- a/test/decompyle/test_global.py
+++ b/test/decompyle/test_global.py
@@ -14,11 +14,11 @@ def a():
         def c():
             k = 34
             global i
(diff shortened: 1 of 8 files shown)


rocky:
Ah - this is what you meant. Ok - will wait until it passes CI and then will merge. 

Probably the same thing is applicable over at the python3-decompile project.

PR #382, opened 2022-01-09
Fix version check comparison

lostbeta:
self.version here is a tuple, hence an exception is raised when comparing it to 3.6 which is a float


rocky (approved):

rocky:
Thanks!

PR #388, opened 2022-04-09
Fix parse37base.py local variable 'v' referenced before assignment

djerryz:
1. Fix Exception:
line 472, in customize_grammar_rules
    "BUILD_SLICE value must be 2 or 3; is %s" % v
UnboundLocalError: local variable 'v' referenced before assignment
2. define v in a better place

### DIFF
--- a/uncompyle6/parsers/parse37base.py
+++ b/uncompyle6/parsers/parse37base.py
@@ -196,7 +196,10 @@ class Python37BaseParser(PythonParser):
             opname_base = opname[: opname.rfind("_")]
 
             # The order of opname listed is roughly sorted below
-
+            try:
+                v = token.attr
+            except:
+                v = "token.attr miss"
             if opname == "LOAD_ASSERT" and "PyPy" in customize:
                 rules_str = """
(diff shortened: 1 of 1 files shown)


rocky:
Thanks for the bug report and fix. As I look at this, the problem with kind of solution is that sets a variable (and here something with a not very descriptive name `v` ) when it is meaningless. 

Adding `try` except isn't the kind of programming construct that should be used in Python programs when it is necessary. Here, there are two alternatives.

Fist, one can know whether there is an `attr` by the opcode type: opcodes below value `HAVE_ARGMENT` do now have an operand and do not have `attr`. 

The second approach in 83ab853 is to just not use `v` here but use `token.attr` which I prefer because it is more specific. 

In general, as a programming principle, I think it better programming practice to set variables only when they are needed rather than set a generic name like `v` that is used sometimes and sometimes not, especially in a function this long and complex. 

I realize though that this is something of a matter of test. (If `v` were named `operand` which is a clearer name, in my opinion that would have been a stronger reason to set it once. But here I would have only set it if it had been used.)

djerryz:
Thanks. Good Job.

Issue #411, opened 2022-09-30
`JUMP_FORWARD` decompilation error


Berbe:
# Description

```python
# uncompyle6 version 3.9.0a1
# Python bytecode version base 2.7 (62211)
# Decompiled from: Python 3.9.2 (default, Feb 28 2021, 17:03:44)
# [GCC 10.2.1 20210110]
# Embedded file name: test.5.KO.py
# Compiled at: 2022-10-01 00:49:19


class Test:

    def test--- This code section failed: ---

 L.   3         0  SETUP_LOOP           47  'to 50'
                3  LOAD_FAST             1  'a'
                6  LOAD_ATTR             0  'values'
                9  CALL_FUNCTION_0       0  None
               12  GET_ITER
               13  FOR_ITER             33  'to 49'
               16  STORE_FAST            2  'b'

 L.   4        19  LOAD_FAST             2  'b'
               22  POP_JUMP_IF_FALSE    13  'to 13'

 L.   5        25  LOAD_FAST             2  'b'
               28  STORE_FAST            1  'a'
               31  JUMP_FORWARD          3  'to 37'

 L.   7        34  CONTINUE             13  'to 13'
             37_0  COME_FROM            31  '31'

 L.   8        37  LOAD_FAST             2  'b'
               40  POP_JUMP_IF_FALSE    13  'to 13'

 L.   9        43  CONTINUE             13  'to 13'
               46  JUMP_BACK            13  'to 13'
               49  POP_BLOCK
             50_0  COME_FROM             0  '0'

Parse error at or near `LOAD_FAST' instruction at offset 37
```

## How to Reproduce

<details>
<summary>Code</summary>

```python
class Test():
    def test(self):
        for b in a.values():
            if b:
                a = b
            else:
                continue
            if b:
                pass
```
</details>

## Output Given

<details>
<summary>Output</summary>

```console
-- Stacks of completed symbols:
START ::= |- stmts .
_come_froms ::= \e__come_froms . COME_FROM
_ifstmts_jump ::= \e_c_stmts_opt . JUMP_FORWARD come_froms
_ifstmts_jump ::= c_stmts_opt . JUMP_FORWARD come_froms
_ifstmts_jump ::= c_stmts_opt JUMP_FORWARD . come_froms
_stmts ::= _stmts . stmt
_stmts ::= stmt .
and ::= expr . JUMP_IF_FALSE_OR_POP expr COME_FROM
[...]

Issue #408, opened 2022-09-20
`JUMP_ABSOLUTE` decompilation error


Berbe:
# Description

I am running against a `JUMP_ABSOLUTE` error with a piece of bytecode written in Python 2.7, using uncompyle6 3.9.0a1 (source code from GitHub, current master branch):
```python
class Test:
    elements = []

    def test--- This code section failed: ---

 L.   4         0  SETUP_LOOP           61  'to 64'
                3  LOAD_GLOBAL           0  'Test'
                6  LOAD_ATTR             1  'elements'
                9  GET_ITER
               10  FOR_ITER             50  'to 63'
               13  STORE_FAST            1  'element'

 L.   5        16  LOAD_FAST             1  'element'
               19  LOAD_ATTR             2  'a'
               22  POP_JUMP_IF_FALSE    10  'to 10'

 L.   6        25  LOAD_FAST             1  'element'
               28  LOAD_ATTR             3  'b'
             31_0  COME_FROM            22  '22'
               31  POP_JUMP_IF_FALSE    10  'to 10'

 L.   7        34  JUMP_FORWARD          3  'to 40'

 L.   9        37  CONTINUE             10  'to 10'
             40_0  COME_FROM            34  '34'

 L.  10        40  LOAD_FAST             1  'element'
               43  LOAD_ATTR             4  'c'
               46  POP_JUMP_IF_FALSE    60  'to 60'

 L.  11        49  LOAD_GLOBAL           5  'Hello'
               52  PRINT_ITEM
               53  PRINT_NEWLINE_CONT
               54  JUMP_ABSOLUTE        60  'to 60'
               57  JUMP_BACK            10  'to 10'
               60  JUMP_BACK            10  'to 10'
               63  POP_BLOCK
             64_0  COME_FROM             0  '0'

Parse error at or near `JUMP_ABSOLUTE' instruction at offset 54
```

## How to Reproduce

<details>
<summary>Code</summary>

```python
class Test():
    elements = []
    def test(self):
        for element in Test.elements:
            if element.a:
                if element.b:
                    pass
                else:
                    continue
                if element.c:
                    print Hello
```
</details>

## Output Given

<details>
<summary>Output</summary>

```console
# uncompyle6 version 3.9.0a1
# Python bytecode version base 2.7 (62211)
# Decompiled from: Python 3.9.2 (default, Feb 28 2021, 17:03:44)
# [GCC 10.2.1 20210110]
# Embedded file name: ./test.3.py
[...]

PR #413, opened 2022-10-06
Fix: CONTINUE in else block in a for loop

Berbe:
Attempt at fixing #411


Berbe:
One immediate problem I got with this is the original `pass` token get decompiled as `continue`.
Since `pass` represent by essence the lack of instruction, it might be possible it is only impossible to recover it. I wonder if the decompiled code _really_ is equivalent to the original.
Test cases of my own:
1. ```python
   class Test():
       def test(self):
           for b in a.values():
               if b:
                   a = b
               else:
                   continue
               if b:
                   pass
   ```
   gets decompiled as:
   ```python
   class Test:

       def test(self):
           for b in a.values():
               if b:
                   a = b
               else:
                   continue
               if b:
                   continue
   ```
2. ```python
   class Test():
       def test(self):
           for b in a.values():
               if b:
                   pass
               else:
                   continue
               if b:
                   a = b
   ```
   gets decompiled to:
   ```python
   class Test:

       def test(self):
           for b in a.values():
               if b:
                   continue
               else:
                   continue
               if b:
                   a = b
   ```

uncompyle6/parsers/reducecheck/ifelsestmt.py
```diff
@@ -240,6 +240,21 @@ def ifelsestmt(self, lhs, n, rule, tree, tokens, first, last):
...
+                  and tree[3][0][0].kind == 'continue'
+                  and if_condition[1][0].attr == tree[3][0][0][0].attr):
+                    # for ...:
+                    #   if ...:
+                    #     ...
+                    #   else:
+                    #     continue
+
+                    return False
+
```
rocky, line 257: This is good! There are a number of small changes though that would improve readability. 

Things like `tree[3]` or worse `tree[3][0][0]` hard to follow. Previously `tree[3]` was assigned to the variable `else_suite` and the variable  `if_condition[1]` was assigned to the variable `jmp`. 

Also we've changed `==` of an ASTnode so that uses `.kind` in for testing. So `tree[2] == "JUMP_FORWARD"` is the same as `tree[2].kind == "JUMP_FORWARD"`.  Note that in new code, I use black conventions which puts strings in double quotes when that is easily possible as it is here. 

Given the above, in general the pattern I have been using is:

```
hoped_for_nonterminal_name = tree[i]
if tree[i] == "hoped_for_nonterminal_name": 
    # other tests and code
```

Missing in the above is the comment general comment mentioned in #411 in words what the intent is. Also missing is the check that the nonterminal we want to reduce on is the kind of "if" that appears inside a loop statement. 

So given all of this here is how I'd write the above: 

```
            if first_offset > jmp_target:
                # A backward or loop jump from the end of an "else"
                # clause before the beginning of the "if" test is okay
                # only if we are trying to match or reduce an "if"
                # statement of the kind that can occur only inside a
                # loop construct.
                if lhs in ("ifelsestmtl", "ifelsestmtc"):
                    jump_false = jmp
                    if (
                        tree[2].kind == "JUMP_FORWARD"
                        and jump_false == "jmp_false"
                        and len(else_suite) == 1
                    ):
                        suite_stmts = else_suite[0]
                        continue_stmt = suite_stmts[0]
                        if (
                            suite_stmts == "suite_stmts"
                            and len(suite_stmts) == 1
                            and continue_stmt == "continue"
                            and jump_false[0].attr == continue_stmt[0].attr
                        ):
                            # for ...:
                            #   if ...:
                            #     ...
                            #   else:
                            #     continue
                            return False
                return True
```

Do you mind revising the above?

rocky:
> One immediate problem I got with this is the original `pass` token get decompiled as `continue`. Since `pass` represent by essence the lack of instruction, it might be possible it is only impossible to recover it. I wonder if the decompiled code _really_ is equivalent to the original. Test cases of my own:
> 
> 1. ```python
>    class Test():
>        def test(self):
>            for b in a.values():
>                if b:
>                    a = b
>                else:
>                    continue
>                if b:
>                    pass
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    gets decompiled as:
>    ```python
>    class Test:
>    
>        def test(self):
>            for b in a.values():
>                if b:
>                    a = b
>                else:
>                    continue
>                if b:
>                    continue
>    ```
> 2. ```python
>    class Test():
>        def test(self):
>            for b in a.values():
>                if b:
>                    pass
>                else:
>                    continue
>                if b:
>                    a = b
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    gets decompiled to:
>    ```python
>    class Test:
>    
>        def test(self):
>            for b in a.values():
>                if b:
>                    continue
>                else:
>                    continue
>                if b:
>                    a = b
>    ```

In the test I tried with the revised patch I don't see this. Please try it yourself and let me know if you still see the "continue" after "if b" or not. 

Also, I find it helpful to really reduce the code to the smallest bytecode. For example, do we really need the class definition in there or do things fail without it?  I am pretty sure `a.values()` an be replace with `a` without any loss but with smaller bytecode.

rocky:
One final thing. In my PR #408 we should add a runnable test for these constructs. I can give more details when we get closer to this being more fully done.

uncompyle6/parsers/reducecheck/ifelsestmt.py
Berbe: >                 if lhs in ("ifelsestmtl", "ifelsestmtc"):

This makes the code fail: `lhs` is `ifelsestmt` both times this code section is attempted, as discussed in the issue.

Berbe: >                 if lhs in ("ifelsestmtl", "ifelsestmtc"):

This makes the code fail: `lhs` is `ifelsestmt` both times this code section is attempted, as discussed in the issue.

rocky: Looks like the "ifelsestmtc" isn't hooked into stmts, in Python 2.7. So this patch hooks it in:
[for-else-continue-2.4.txt](https://github.com/rocky/python-uncompyle6/files/9736476/for-else-continue-2.4.txt)

Berbe:
I simplified my tests.
Note my test `2.` supplied earlier was wrong: I put a `pass` instruction where a `continue` was required for decompilation to fail. Cf. comment below.

1. ```python
   for b in a:
       if b:
           a = b
       else:
           continue
       if b:
           pass
   ```
2. ```python
   for b in a:
       if b:
           #pass # Works
           continue
       else:
           continue
       if b:
           a = b
   ```

Test `1.` gets decompiled to:
```python
for b in a:
    if b:
        a = b
    else:
        continue
    if b:
        continue
```

rocky:
> I simplified my tests. Note my test `2.` supplied earlier was wrong: I put a `pass` instruction where a `continue` was required for decompilation to fail. Cf. comment below.
> 
> 1. ```python
>    for b in a:
>        if b:
>            a = b
>        else:
>            continue
>        if b:
>            pass
>    ```
> 2. ```python
>    for b in a:
>        if b:
>            #pass # Works
>            continue
>        else:
>            continue
>        if b:
>            a = b
>    ```
> 
> Both tests pass, code failing to decompile on current `master` and works with the contents of this PR. However, test `1.` gets decompiled to:
> 
> ```python
> for b in a:
>     if b:
>         a = b
>     else:
>         continue
>     if b:
>         continue
> ```

Here, isn't `continue` here is the same thing as `pass`?  

Should one want to distinguish the two, this is where using  peculiarities of code that you had remarked is weird or has dead code could be used to distinguish. Personally, I am not sure I'd want to go down that path though. 

Ok. So now the next step is to write very simple program that when run just terminates normally but will fail or raise an assertion when the program decompiiles incorrectly or run through the current master . See https://github.com/rocky/python-uncompyle6/blob/master/test/simple_source/bug27%2B/06_raise.py for an example of such a program that was written in the past. 

There is one small addition in current convention; add the docstring: 
```
"""This program is self-checking!"""
```
As the first executable line after the comments. Previously I was using the comment:
```
# RUNNABLE!
```
but of course that gets stripped out of the final bytecode. The docstring however does not get removed unless an optimization level is used in byte-compiling the code. 

After that is done then, add this to the git repository and byte compile the program using `test/add-test.py` supplying option `--run`.  Here though you should make sure that when you run this it is run with a Python 2.7 interpreter so we get Python 2.7 bytecode. For example.py if the test program is put in `test/simple_source/bug27+/04_for_if_else_continue.py`, then from inside the `test` directory you would run:

```
$ python2.7 ./add-test.py --run simle_source/bug27/04_for_if_else_continue.py
$ git add --force bytecode_2.7_run/04_for_if_else_continue.pyc
```

That last git command is to make sure the bytecode gets into the repository. And it has to have the `--force` option because bytecode files or those files ending in pyc are ignored by default.

Finally, after this is done you can check that this test got added, again from the test directory, by running: 

```
make check-bytecode-2.7
```

And that's it!

Berbe:
I can't seem to be able to run the `uncompyle6` package in a Python 2.7 venv.
What I was doing was compiling code in a Python 2.7 venv and uncompiling with `uncompyle6` in a Python 3.

Compiling tests with your script seems to require the former and I'm beat on that one.

rocky:
> I can't seem to be able to run the `uncompyle6` package in a Python 2.7 venv. What I was doing was compiling code in a Python 2.7 venv and uncompiling with `uncompyle6` in a Python 3.
> 
> Compiling tests with your script seems to require the former and I'm beat on that one.
> 
> Is this OK if I compile the test separately (not going through your script) in a Python 2.7 venv and simply add source & bytecode to the repository?

That is perfectly fine!   (The code for Python 2.7 is in the python-2.4 branch. And you'd need to use code from the python-2.4 branch of xdis and spark-parser)

Berbe:
Alright I added a test file and its 2.7 compiled version.
Could you check if this runs fine in the pipeline?

uncompyle6/parsers/reducecheck/ifelsestmt.py
```diff
@@ -240,6 +259,34 @@ def ifelsestmt(self, lhs, n, rule, tree, tokens, first, last):
...
+                    if (
+                        tree[2] == "JUMP_FORWARD"
+                        and jump_false == "jmp_false"
+                        and len(else_suite) == 1
+                    ):
+                        suite_stmts = else_suite[0]
+                        continue_stmt = suite_stmts[0]
+                        if(
+                            suite_stmts == "suite_stmts"
+                            and len(suite_stmts) == 1
```
rocky, line 278: There has been something nagging me about this. I suspect that rather than checking that there is only _one_ statement and it is `continue` we should be checking on the _last_ statement  being `continue`.

Berbe: `tokens[last]`?

test/simple_source/bug27+/07_for_if_else-continue.py
```diff
@@ -0,0 +1,27 @@
...
+def test1():
+    a = []
+    for b in a:
+        if b:
+            a = b
+        else:
+            continue
+        if b:
+            pass
+
```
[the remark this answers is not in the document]
Berbe: Updated tests: let me know.

uncompyle6/parsers/reducecheck/ifelsestmt.py
Berbe: > Looks like the "ifelsestmtc" isn't hooked into stmts, in Python 2.7. So this patch hooks it in: [for-else-continue-2.4.txt](https://github.com/rocky/python-uncompyle6/files/9736476/for-else-continue-2.4.txt)

Your changes change drastically what is happening: Now the change I made in `ifelsestmt()` is rendered useless:
```python
[...]
L. 12:  19-37  ifelsestmtc ::= testexpr c_stmts_opt JUMP_FORWARD else_suite come_froms (14)
L. 12:  19-37  ifelsestmt ::= testexpr c_stmts_opt JUMP_FORWARD else_suite come_froms (14)
Reduce ifelsestmt invalid by check
L. 12:  19     stmt ::= ifelsestmtc (14)
[...]
```

rocky: > ```python
> Reduce ifelsestmt invalid by check
> ```

I don't see a problem with this. `ifelsestmt` is invalid here. Instead `ifelsestmtc` should match instead. 

If there is a specific problem with decompilation, create a little test case that shows this.

rocky: > Since your fix resolved the issue, shouldn't yo uauthor the changes rather than me? You can pick my test files up if you like for integration tests.

With respect to who is marked on fixing for this particular issue, I don't really care. My goal here was to get you more educated on this and this was a faster way to show you how things work.

Berbe: > > ```python
> > Reduce ifelsestmt invalid by check
> > ```
> 
> I don't see a problem with this. `ifelsestmt` is invalid here. Instead, `ifelsestmtc` should match.
> 
> If there is a specific problem with decompilation, create a little test case that shows this.

I made the test by reverting my changes, hence nullifying the changes I brought in this PR.
You patch did all the work and both my test cases decompile.

test/simple_source/bug27+/07_for_if_else-continue.py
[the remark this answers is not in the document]
rocky: Tests are better. 

```
assert r == True, "Incorrect flow"
```

is a shorter and more natural way to write: 

```
if r != True:
    AssertionError("Incorrect flow")
```

In fact it is better because when there is a failure we should see the failure on the test1 or test2 call rather than a couple lines down from that.

uncompyle6/parsers/reducecheck/ifelsestmt.py
rocky: I don't mind giving you credit for work I did, but if you feel bad about this, I'll push a new branch with the changes. And you can rebase this PR on top of that. Let me know if that's what you want me to do.

uncompyle6/parsers/reducecheck/ifelsestmt.py
rocky: Somewhere or at sometime you mentioned using `tokens[last]` to look for `continue`. 

While there are many ways to do things. A potential problem with `tokens[last]` is that it might have fabricated tokens like "COME_FROM".   From experience, the preferred way to do this would be to use `suite_stmts.last_child()`.

uncompyle6/parsers/reducecheck/ifelsestmt.py
Berbe: My point was: there is no need to modify the function I was trying to change anymore.
The fix is the patch you provided.

I changed authorship for you in the related commit of this PR (I could do that since it appears you do not sign yours with GPG over here).

I changed my additional commit to merely include tests, but I'm yet unsure about their contents.

uncompyle6/parsers/reducecheck/ifelsestmt.py
Berbe: Yeah I removed my comment about `tokens[last]` because it was stupid, thinking about it.
`tokens` contains all the instructions from the sequence; what you meant was getting the last token _from the `else` block_.

I hadn't come across that handy `last_child()` helper function, and I'll be sure to try and remember that when/if I need it! Thanks for that.

test/simple_source/bug27+/07_for_if_else-continue.py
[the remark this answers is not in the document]
Berbe: I rewrote the tests using the new syntax, and it actually helped me identifying a mistake:
```python
r = test1([False])
if r != False:
    AssertionError("Incorrect flow")
```
was wrong and should have errored out, since the expected answer was `None`. This way of writing the test somehow works, meaning `None` is considered equivalent as `False`.
However, the following (new) way of writing the test works as expected:
```python
assert test1([False]) == None, "Incorrect flow"
```
I thought Python was robuster with comparison operators since there is no _strict equal_ one compared to some other languages...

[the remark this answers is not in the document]
rocky: > I rewrote the tests using the new syntax, and it actually helped me identifying a mistake:
> 
> ```python
> r = test1([False])
> if r != False:
>     AssertionError("Incorrect flow")
> ```
> 
> was wrong and should have errored out, since the expected answer was `None`. This way of writing the test somehow works, meaning `None` is considered equivalent as `False`. However, the following (new) way of writing the test works as expected:
> 
> ```python
> assert test1([False]) == None, "Incorrect flow"
> ```
> 
> I thought Python was robuster with comparison operators since there is no _strict equal_ one compared to some other languages...

The stricter check for None is to use `is None and using a lint checker like flake8, or pylint would probably have told you.

While I am not a fan of Python, (see [this](https://qr.ae/pvcKAo) written 5 years ago), it _does_ does allow for stricter checks. 

Javascript is similar is having quirks about type casting, and if I recall, it has even weirder ones for those who don't keep up with Javascript development. 

Back to the issue at hand. With respect to the tests, first, they all fail (I guess because they don't decompile properly). If that is the case then, look at the bytecode produced and the parse tree produced (with `--tree`) to make sure that the code of the tests uses the new reductions and not some others. 

I'd also try adding statements before the else continue to see of that works. If so, then the test can and should be changed to leave a bread-crumb that that part of the "if" or "else" was taken and then that can be checked on the calling side.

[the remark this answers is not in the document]
Berbe: > While I am not a fan of Python, (see [this](https://qr.ae/pvcKAo) written 5 years ago), it _does_ does allow for stricter checks.

Surprising, coming from someone maintaining a decompiler for it! I would have naturally guessed some degree of affinity with the language.

I don't know if you know a lot about PHP, but this language has an "interesting" feature: its comparison operators are [circular](http://www.phpsadness.com/sad/52)! You're in for a good laugh.

---
I corrected the comparison operator with `is` for `None`

---
I find grammar display (`-g`) handier to identify how the code is parsed:
```console
{...]
L.  7:  13-43  ifelsestmt ::= testexpr c_stmts_opt JUMP_FORWARD else_suite come_froms (16)
Reduce ifelsestmt invalid by check
[...]
L. 18:  13-52  ifelsestmt ::= testexpr c_stmts_opt JUMP_FORWARD else_suite come_froms (19)
Reduce ifelsestmt invalid by check
```

rocky:
Sorry for the delay. We'll go with this, but I suspect there is more work to do. 

In terms of work that _could_ be done if there were interest is matching whether `continue` or `pass` was used based on the form of instructions that get emitted. That however would best be done in the tree transformation phase.

PR #415, opened 2022-11-05
Add missing ParserError2 in make_functions36

jameshilliard:
Should fix:
```python
uncompyle6/semantics/make_function36.py", line 170, in make_function36
    except (ParserError, ParserError2) as p:
NameError: name 'ParserError2' is not defined
```


rocky:
Thanks.

PR #416, opened 2022-11-05
Fix ImportError: cannot import name PYTHON_VERSION

jameshilliard:
Seems these didn't get updated.


rocky:
Thanks. (They are mostly in-line standalone tests).

PR #418, opened 2022-11-05
Fix status_msg args

jameshilliard:
Fixes:
```python
uncompyle6/bin/uncompile.py", line 201, in main_bin
    mess = status_msg(do_verify, *result)
TypeError: status_msg() takes 5 positional arguments but 6 were given
```


rocky:
LGTM - thanks

PR #419, opened 2022-11-05
Replace remaining self.opc with token.opc

jameshilliard:
Seems these were missed in 207dc0b506a60fdc3cc22264733f79d95ebcfc83.


rocky:
Probably right, but waiting to see a concrete example as an issue.

Issue #437, opened 2023-04-08
Wrong indent on Python 2.7


andrem-eberle:
I am running uncompyle6 on this code:
```
if a == 2:
	print(a)
else:
	if a == 4:
		print(4)		
	else:
		if a == 6:
			print(6)
		else:
			if a == 7:
				print(7)
			else:
				pass
	print(3)

def somefunc():
	print('somefunc')

try:
	t = None
except Exception:
	t = {}
```
After compileall & uncompyle6 I get this:

```
if a == 2:
    print a
else:
    if a == 4:
        print 4
    else:
        if a == 6:
            print 6
        elif a == 7:
            print 7
        print 3

    def somefunc():
        print 'somefunc'


    try:
        t = None
    except Exception:
        t = {}
```

Later part (from function onwards) has wrong indent

andrem-eberle:
I rummaged a bit through the source code, and did some tweak to the grammar rules which seems to have fixed this. Catching all blocks in the flow with just rules is quite tricky. I ran the tests in ./test and seems nothing was broken. I sent a pull request for you to evaluate.

rocky:
Fixed by #440

PR #438, opened 2023-04-08
In reference to issue #437, I did some changes in the grammar rules, …

andrem-eberle:
…particularly changed a few ifelsestmt in 2.7 to accept a single COME_FROM at the end. It fixed that issue. I ran the tests on ./test and nothing was broken so far.

https://github.com/rocky/python-uncompyle6/issues/437

### DIFF
--- a/uncompyle6/semantics/customize36.py
+++ b/uncompyle6/semantics/customize36.py
@@ -207,17 +207,12 @@ def customize_for_version36(self, version):
         elif build_class[1][0] == "load_closure":
             # Python 3 with closures not functions
             load_closure = build_class[1]
-            if load_closure[-4] == "LOAD_CODE":
-                subclass_code = load_closure[-4].attr
-            elif hasattr(load_closure[-3], "attr"):
-                # Python 3.3 classes with closures work like this.
-                # Note have to test before 3.2 case because
-                # index -2 also has an attr.
-                subclass_code = load_closure[-3].attr
-            elif hasattr(load_closure[-2], "attr"):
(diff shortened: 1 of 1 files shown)


rocky:
Thanks for digging into this and submitting a PR. Alas, this kind of approach I have found is not the way to go (circa 2016 I would have probably accepted this).

This kind of change is too heuristic, like ChatGPT. (And the similarity of the two leads me to believe a strictly heuristic to learning this way will never give  you totally accurate results  which is what is what you are asking for here.)

Since back then the better approach is to check that the rule is correct in a reduction rule. See https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/reducecheck/ifelsestmt.py and
https://github.com/rocky/python-uncompyle6/blob/master/uncompyle6/parsers/reducecheck/ifelsestmt2.py

What we'd want to check is that the jump inside `testexpr` jumps exactly to the end of the `come_froms`. If it turns out that `come_froms` gobbles up too much, we can make use of the ambiguity allowed by the context-free grammar scheme and add also that single COME_FROM as you have in this PR.

> …particularly changed a few ifelsestmt in 2.7 to accept a single COME_FROM at the end. It fixed that issue. I ran the tests on ./test and nothing was broken so far.

The tests aren't that good. A more complete set of tests is kicked off in test/stdlib/runtests.sh. 

And to run that you need some version of Python 2.7 available and the `python-2.4` branch checked out.

andrem-eberle:
I see, I might look further into this later.

Issue #439, opened 2023-04-08
Wrong behaviour on long lists in Python 2.x


andrem-eberle:
## Description

Decompiled result incorrect for certain types of array:

```
print ['y', Exception, 'x',Exception, 'z']
```

## How to Reproduce

Input is a file with this line:
```
print(['y',Exception,'x',Exception,'z'])
````

```
python -m compileall file.py
uncompyle6 file.pyc
```

## Output Given

```
print ['y', 'Exception', 'x', 'Exception', 'z']
````

It transforms the global Exception into a string. The --asm output gives me this (relevant part):
```
0_0  COLLECTION_START      0  'CONST_LIST'

 L.   1         0  ADD_VALUE             0  'y'
                3  ADD_VALUE             0  'Exception'
                6  ADD_VALUE             1  'x'
                9  ADD_VALUE             0  'Exception'
               12  ADD_VALUE             2  'z'
               15  BUILD_CONST_LIST      5  None

```

## Expected behavior

Globals should be decompiled as global names, not as strings

## Environment

uncompyle6 3.9.1.dev0
Using Python 2.7.18 for compileall and Python 3.8.16 for uncompyle6
Ubuntu 16.x

andrem-eberle:
I have been looking into this. Some easy solution I found was to wrap the pattr in scanner2 and set repr to give a version without quotes for the NAME ops. Tell me if that is an acceptable solution and I will prepare a pull request.

rocky:
> I have been looking into this. Some easy solution I found was to wrap the pattr in scanner2 and set repr to give a version without quotes for the NAME ops. Tell me if that is an acceptable solution and I will prepare a pull request.

I think this would be okay if I understand correctly 

In instructions and tokens, I believe we have access to the underlying type and that could help to indicate whether there should be quotes or not.

rocky:
This situation is very specific to the pseudo opcode  `ADD_VALUE` and issues around the tokens. (`ADD_VALUE` and `COLLECTION_START` are necessary to speed up parsing which is yet another long-running weakness of the code).

A proposed solution is to change the Token object to indicate more operand type information. But right now, the only information we need to add is whether an operand is a constant string or an identifier of some sort. In particular the string "Exception" versus the name Exception.

Operands of `LOAD_NAME` are never strings but are names. The only `LOAD_.`.. operands that can be strings is those of `LOAD_CONST` sometimes. If the operat and is not a code type and possibly some other types, the operand is a string.

[...]

PR #440, opened 2023-04-08
Tentative fix for issue #437.

andrem-eberle:
I followed your guidelines, and reworked ifelsestmt.py a bit. I just added another set of rules that is then checked for offsets, it fixed this particular issue (#437) at least. I know the tests in test/ aren't great, but at least they aren't failing :)

I ran the tests on test/stdlib/runtests.sh also (Python 2.7) but I keep getting an error, which I included in issue #439. I might look into that one also since I am working on it.


rocky:
LGTM - this was the kind of change I was expecting in the back of my mind. Thanks!

PR #445, opened 2023-04-14
Tentative fix for issue #439

andrem-eberle:
What I did was add a wrapper to pattr in scanner2.py, when transforming the disasm into the intermediate abstraction. It is a class, that essentially gives a specific __repr__ without the quotes when the translator emits the source code later. I added a few str() around to make sure the behavior is consistent overall.

Tests in test/ worked, still struggling with test/stdlib ones, but I am trying to fix all the unicode stuff.

### DIFF
--- a/uncompyle6/scanners/scanner2.py
+++ b/uncompyle6/scanners/scanner2.py
@@ -360,7 +360,7 @@ class Scanner2(Scanner):
                         pattr = const
                         pass
                 elif op in self.opc.NAME_OPS:
-                    pattr = names[oparg]
+                    pattr = PattrWrapper(names[oparg])
                 elif op in self.opc.JREL_OPS:
                     #  use instead: hasattr(self, 'patch_continue'): ?
                     if self.version[:2] == (2, 7):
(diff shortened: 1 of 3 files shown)


andrem-eberle:
I will look into the failed tests

andrem-eberle:
I have been going through the tests, it seems the automated tests are expecting quotes around the global names. I am wondering what is the best way to deal with this. Should I add quotes on the --asm flag, or rework the tests?

rocky:
@andrem-eberle I spent a little time looking at this particular problem. 

This situation is very specific to the pesudo instruction `ADD_VALUE` and issues around the tokens. (`ADD_VALUE` and `COLLECTION_START` are necessary to speed up parsing which is yet another long-running weakness of the code).

A proposed solution is to change the Token object to indicate more operand type information. But right now, the only information we need to add is whether an operand is a constant string or an identifier of some sort. In particular the string "Exception" versus the name Exception. 

Operands of `LOAD_NAME` are never strings but are names. The only `LOAD_...`  operands that can be strings is those of `LOAD_CONST` _sometimes_.  If the operat and is not a code type and possibly some other types, the operand is a string.

Possibly this is something that would be made simpler by better disambiguation in the instruction object inside xdis.

I was going to go down the path of fixing this, but maybe you'd like to investigate and come up with a solution.

PR #452, opened 2023-04-18
Proposed fix for the extra quotes

andrem-eberle:
I separated repr and str by Python version. The test is still broken, the 3.8 bytecode is being decompiled in wrong blocks, I might look into that later.

### DIFF
--- a/uncompyle6/semantics/n_actions.py
+++ b/uncompyle6/semantics/n_actions.py
@@ -267,7 +267,10 @@ class NonterminalActions:
                 if elem == "add_value":
                     elem = elem[0]
                 if elem == "ADD_VALUE":
-                    value = "%r" % elem.pattr
+                    if self.version[0] == 2:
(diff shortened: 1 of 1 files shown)


uncompyle6/semantics/n_actions.py
```diff
@@ -267,7 +267,10 @@ def n_const_list(self, node):
                 if elem == "add_value":
                     elem = elem[0]
                 if elem == "ADD_VALUE":
-                    value = "%r" % elem.pattr
+                    if self.version[0] == 2:
```
rocky, line 270: There is also Python 1.x. The convention used so far is `self.version < (3, 0)` which I think is clearer.

rocky:
Do you know if the breakage is _exactly_ between Python 2.7 and 3.0 or could it between some other two successive versions?

rocky:
I will make the changes I suggest and revise the test in a future chagne.

PR #453, opened 2023-04-18
Support high version python to decompile low version pyc

JANlittle:
Uncompyle6 uses the current Python interpreter version as the default .pyc version, which prevents Python 3.11 from working, even if the .pyc version is 3.8 or lower. I made some small modifications to specify the version of .pyc when generating tokens, which may be helpful. Due to time constraints, I only tested a few pyc file. Sorry for that. :(
Also, it maybe need to change `xdis/std.py`:
```python
from uncompyle6 import PYC_version
_std_api = make_std_api(PYC_version)
```

### DIFF
--- a/uncompyle6/__init__.py
+++ b/uncompyle6/__init__.py
@@ -46,6 +46,9 @@ from uncompyle6.main import decompile_file  # noqa
 
 from uncompyle6.semantics.pysource import code_deparse, deparse_code2str
 
+
+PYC_version = 3.6
+
 __all__ = [
     "__version__",
     "code_deparse",
--- a/uncompyle6/bin/uncompile.py
+++ b/uncompyle6/bin/uncompile.py
@@ -62,7 +62,7 @@ Extensions of generated files:
 
 program = 'uncompyle6'
 
-from uncompyle6 import verify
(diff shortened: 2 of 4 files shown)


rocky:
I don't understand. I can build uncompyle6 from 3.9 and when I pass it a bytecode file from 3.0 it works. 

```
$ ./bin/uncompyle6 test/bytecode_3.0/00_import.pyc 
# uncompyle6 version 3.9.1.dev0
# Python bytecode version base 3.0 (3131)
# Decompiled from: Python 3.9.16 (main, Dec 13 2022, 19:28:33) 
# [GCC 11.3.0]
# Embedded file name: simple_source/stmts/00_import.py
# Compiled at: 2019-07-10 15:49:35
import sys
from os import path
from os import *
import time as time1, os as os1, http.client as httpclient
if len(__file__) == 0:
    import a.b.c as d, stuff0.stuff1.stuff2.stuff3 as stuff3
# okay decompiling test/bytecode_3.0/00_import.pyc
```

So what exactly is it that you want to do? 

It feels like the only thing this may do is allow one not to have to specify a version somewhere, and I don't see this that important.

PR #458, opened 2023-05-31
test: Expose double return bug

Et7f3:
This PR only add test.


rocky (changes requested):
Obviously we cannot commit something that breaks things without the fix. 
The source code for this should appear in `test/simple_source` somewhere.

Et7f3:
> Obviously we cannot commit something that breaks things without the fix

I put in draft to avoid mistake. Other repo propose this so a dev could directly push to that branch and get the CI green.

>  The source code for this should appear in test/simple_source somewhere.

Thanks I didn't saw it.

PR #505, opened 2024-11-12
Use a single TABLE copy

gdesmar:
Do not overwrite possible previous updates to the new TABLE_R and TABLE_DIRECT dictionaries.


rocky:
LGTM

